PR #36, opened 2015-12-20
Added a link to the expect definition file

andrew-w-ross:


blakeembrey:
Awesome :+1: You're just missing one thing, and that's to specify `main` in `typings.json`. That's what the build is complaining about and I'll definitely improve that error ASAP.

blakeembrey:
Oh, I just realised. Is this meant to be an ambient module? I suspect you want to make it external (https://github.com/andrew-w-ross/typings-expect/blob/master/expect.d.ts) by pulling everything that's inside the ambient declaration outside and removing it.

andrew-w-ross:
Oh it's not supposed to be an ambient module. What have I done wrong?

blakeembrey:
Looks like you managed to fix it up :+1: Cheers!

PR #38, opened 2015-12-24
Expect update and readme update

andrew-w-ross:
I’ve updated the readme to mention there is a test function and you should use it before doing a pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 1. Write a typings definition and upload to GitHub - here's an [example using debug](https://github.com/typings/typed-debug/blob/master/typings.json)
 2. Use [semver](http://semver.org/) to specify which version the typing is valid for - here's [debug again](https://github.com/typings/registry/blob/master/npm/debug.json)
 3. Run `npm test` and check that the registry is valid, once passed do a pull request
-3. Once the pull request is merged, it is used as the canonical reference - `typings install debug`
+4. Once the pull request is merged, it is used as the canonical reference - `typings install debug`
 
 ## Structure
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,7 +5,8 @@
 ## Contributing
 
 1. Write a typings definition and upload to GitHub - here's an [example using debug](https://github.com/typings/typed-debug/blob/master/typings.json)
-2. Make a pull request, using [semver](http://semver.org/) to specify which version the typing is valid for - here's [debug again](https://github.com/typings/registry/blob/master/npm/debug.json)
+2. Use [semver](http://semver.org/) to specify which version the typing is valid for - here's [debug again](https://github.com/typings/registry/blob/master/npm/debug.json)
+3. Run `npm test` and check that the registry is valid, once passed do a pull request
 3. Once the pull request is merged, it is used as the canonical reference - `typings install debug`
```
blakeembrey: Would up mind updating this to `4.` now?

blakeembrey:
Thanks, just the one comment :smile:

README.md
andrew-w-ross: I'm really confused why it looks like this in the diff. I have nothing to commit or sync and this is how [my one looks like](https://github.com/andrew-w-ross/registry).

blakeembrey: That's just how markdown is parsed. It only has to start with a number, they could be 1. 1. 1. 1. and it'll still render correctly because they become `<li>` elements which can't actually set the number. This is the raw Markdown though, if you could just quickly correct that :smile:

andrew-w-ross: I see, thanks blake. Merry x-mas btw.

blakeembrey:
Thanks for the fix :+1: Merry Christmas to you too :smile:

PR #44, opened 2016-01-05
update readme

samvloeberghs:
Hi @blakeembrey, @basarat, @gdi2290,

This is my first try at the project. I have made some changes and have some questions:

- I've updated the readme slightly to also include "run npm install". Something I always forget :) Ask me to remove it if you don't like it and I'll commit it in this pull request

- I've added .idea in the .gitignore as I use webstorm and others might as well

- Is it possible to use 6.1.x in the semver? The typings i've created are for 6.1.x in general. Or should I than just use 6.1?

- Can you validate it? I might do some more work in merging more stuff into here if I get the hang on how to do it the right way!

Grtz and good job on the project!

### DIFF
--- a/ambient/foundation-sites.json
+++ b/ambient/foundation-sites.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "6.1.0": "github:samvloeberghs/foundation-sites-typings/index.d.ts#ddfa8f2eff7770d5cbb67c86557a660aee2c8456"
+  }
+}
--- a/ambient/foundation-sites.json
+++ b/ambient/foundation-sites.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "6.1.0": "github:samvloeberghs/foundation-sites-typings/index.d.ts#ddfa8f2eff7770d5cbb67c86557a660aee2c8456"
+    "6.1.0": "github:samvloeberghs/foundation-sites-typings#95104860070bdea179b0506e9aeeafe86440e37e"
   }
 }
--- a/ambient/foundation-sites.json
+++ b/ambient/foundation-sites.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "6.1.0": "github:samvloeberghs/foundation-sites-typings#95104860070bdea179b0506e9aeeafe86440e37e"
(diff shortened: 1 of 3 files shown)


ambient/foundation-sites.json
```diff
@@ -0,0 +1,5 @@
+{
+  "versions": {
+    "6.1.0": "github:samvloeberghs/foundation-sites-typings/index.d.ts#ddfa8f2eff7770d5cbb67c86557a660aee2c8456"
```
blakeembrey: I would just make this `github:samvloeberghs/foundation-sites-typings#ddfa8f2eff7770d5cbb67c86557a660aee2c8456` (without `index.d.ts`) so it'll resolve the `typings.json`. However, since `jquery` is an ambient definition in `typings.json`, you'll need to rename `dependencies` to `ambientDependencies` in `typings.json`. Does that make sense? I'm trying to find an easier way to explain and show this for people that aren't me, so apologies in advance :smile:

blakeembrey:
> I've updated the readme slightly to also include "run npm install". Something I always forget :) Ask me to remove it if you don't like it and I'll commit it in this pull request

Awesome, thanks :+1: 

> I've added .idea in the .gitignore as I use webstorm and others might as well

No worries.

> Is it possible to use 6.1.x in the semver? The typings i've created are for 6.1.x in general. Or should I than just use 6.1?

Right now it's the lowest supported version. It's possible we should play around with this, but I really need to think more about the implications and locations where the version should be stored and how this registry can be better updated to reflect reality. It's likely this version won't be the final one, but for now just use the lowest supported version (they'll get sorted using this in the UI).

Actually, one reason why we use static versions instead of a range is that searching can use a range. Since the `semver` module can't (currently) detect semver range overlaps (E.g. does this range union cause an empty set?) we're sorting on the version field.

> Can you validate it? I might do some more work in merging more stuff into here if I get the hang on how to do it the right way!

I left a comment for you, please don't get overwhelmed :+1: It's a little bit right now, but I'm hoping to make it much easier in time. Hang in there :smile:

samvloeberghs:
@blakeembrey thx for the feedback! I've updated the reference ( see commit https://github.com/typings/registry/commit/89a608635524213294bb1bbfc14e00efbbc9d412 )

blakeembrey:
Thanks for your work! :+1:

PR #50, opened 2016-01-22
feat(ambient): moment

gdi2290:
see issue https://github.com/typings/typings/issues/113 for context


blakeembrey:
This isn't correct. It's only a workaround for an inconvenience. If that's required, we should spec an `overrides/` folder - but that's only needed because DefinitelyTyped is so inconsistent, so I'm hesitant to do that.

From the README:

> ambient for ambient dependencies (typings that describe our programming environment)

The `moment-node` definition in DefinitelyTyped is exactly the type of hacky definition we're meant to be trying to solve.

PR #51, opened 2016-01-22
feat(ambient): zone

gdi2290:
include `zone.js` type


blakeembrey:
Reference: https://github.com/angular/DefinitelyTyped/blob/master/zone/zone.d.ts

I think this one is good, but it'd be cool to pull this out into a typed repo and correct the `es6-shim` dependency to define ambient interfaces.

gdi2290:
cc @alexeagle

alexeagle:
oh, our fork of DT should go away once I get this green: https://github.com/angular/angular/pull/6267
should come from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/zone.js/zone.js.d.ts instead

PR #60, opened 2016-01-26
Added expect-jsx, redux

andrew-w-ross:
I’ve added expect-jsx, redux typings but needed to move expect to an ambient module due to the declaration merging issue typings/typings#119

### DIFF
--- a/ambient/expect-jsx.json
+++ b/ambient/expect-jsx.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.2.2": "github:andrew-w-ross/typings-expect-jsx#d1b640dc7814ddafffa007d86f9d126221756c16"
+  }
+}
--- a/ambient/expect.json
+++ b/ambient/expect.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.13.4": "github:andrew-w-ross/typings-expect#aeceb0cfd309845d4d7d06f7765a360221d0fdba"
+  }
+}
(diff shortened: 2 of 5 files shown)


blakeembrey:
Do you think `import React = __React` should be `import React = require('react')` instead so we aren't tying ourselves to the global that happens to exists?

Aside from that, what's your current testing strategy? Are you just using bundle and `tsc` and checking that it works? That's a high priority thing for me once I get time to fix, so it's interesting to see you've already got something work for you.

blakeembrey:
Also, should Redux be ambient too? Anyway, with your responses, it'll be great to merge - thanks!

andrew-w-ross:
I think you might be right about the `import React = require('react')` let me have a look. 

You're dead right about the testing it's simple case of run tsc and see if it fails. It's why I always `"noEmit":true` on. Overall I try to get the documentation code for the api to compile with as few changes as possible. 

As for redux, the reason it's ambient is because I'm planning on adding support for [redux-thunk](https://github.com/gaearon/redux-thunk) and it will extend the defintion of `IActionCreator`. I'm not a fan of doing it that way but I can't think of anything else. On that note there seems to be [some new changes for typescript 1.8](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#augmenting-globalmodule-scope-from-modules) that might help with this. But I'll leave the declaration merging story up to you.

blakeembrey:
Great, just leave a comment when/if the React change occurs. We'll have a testing tool eventually (2-3 weeks) to avoid that hack (I try to keep the main file representative of the module). As for ambient merging, that's the issue I was seeing - so yeah, not out yet. Looking forward to tweaks like those.

andrew-w-ross:
Ok that's the react change completed.

blakeembrey:
Thanks for much for contributing :+1:

PR #61, opened 2016-01-28
Add npm:domready

unional:
https://github.com/typings/typings/issues/129

### DIFF
--- a/npm/domready.json
+++ b/npm/domready.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.8": "github:unional/typed-domready#881449d638c89897b1e70172ab5413b8886b4ef9"
+  }
+}


blakeembrey:
Awesome, cheers! The failure is complaining about the usage of `browser` field without `main` set. The browser field is an override, you'll still need `main`. In this case, I'd just change `"browser"` to `"main"` and update this PR and we'll be passing :+1:

Reference: https://github.com/unional/typed-domready/blob/master/typings.json

Thanks for this! You're helping identify places where my document must be terrible too.

unional:
Thanks for the feedback. It passes now. :+1:

unional:
I rebase it to your latest code to get rid of the extra merge

blakeembrey:
LGTM :+1: Thanks for contributing.

PR #62, opened 2016-01-28
added tone.js typings

lukephills:


blakeembrey:
Awesome, cheers! Quick question, is everything in the library always global? Aside from that, I found https://github.com/BlokDust/Tone.js/blob/master/utils/TypeScript/Tone.d.ts. If you don't want to maintain it in two places, you're welcome to point it there instead.

E.g. `github:BlokDust/Tone.js/blob/master/utils/TypeScript/Tone.d.ts#ad623816d5a8d9742bde5fe34f12b24d8e5e5f62`.

blakeembrey:
@lukephills Sorry, I had a typo above. I just corrected it.

lukephills:
Oh that would be good to point it there instead. I'm having trouble though. Any idea what i'm missing?

blakeembrey:
Try the URL I updated in the comment above, I had a typo previously (accidentally left `blob/master/` in the path).

lukephills:
OK great thanks! Everything is written globally like this:

var mod = new Tone.Module();
mod.doSomething();

blakeembrey:
LGTM! :+1: Thanks for contributing.

blakeembrey:
Just FYI, in the future if there's a new folder for "global" dependencies, I'll ping you in case I'm going to be moving it about.

lukephills:
Ok great thanks!

PR #71, opened 2016-01-31
typings for pg and generic-pool

goenning:


blakeembrey:
Amazing, thank you! :+1:

PR #84, opened 2016-02-04
Add typed-gulp

unional:
~~Does version `3.8` works?
Original DT file says it is for `Gulp v3.8.x`~~

Ok, Travis answered that question for me. :smile:

### DIFF
--- a/npm/gulp.json
+++ b/npm/gulp.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "3.8.0": "github:typings/typed-gulp#4c4a80e3c7606c5200b0c80885618e8c34622d8f"
+  }
+}


blakeembrey:
Yes, just needs to be the lower semantic version range.

blakeembrey:
Can we avoid merging this until the dependencies are typed too? I'd like to avoid confusing/broken dependency trees existing in the registry.

unional:
Meaning you want `orchestrator` and `node` to be typed first?

blakeembrey:
Just `orchestrator`, that's a dependency. Node is fine as is, that's an ambient dependency (E.g. provided by runtime).

unional:
Ok.

unional:
@blakeembrey , updated dependency. Please check. :smile:

unional:
Fix #78

blakeembrey:
I'm ok with the merge (if you are), but it definitely needs large refactoring to make the type information more useful. Right now you won't be able to access any type information in the exported module.

To make it properly, you'd have to use a namespace or top-level exports.

unional:
Arr... I messed up my `test\tsconfig.json`.  I thought it is working. :sob: 

Should it be something like this?
```ts
namespace gulp {
  export function src() {...}
  export function task() {...}
  ...
}

export = gulp
```

I previously use TS only for ES6 + type system. So I still have a lot to learn on the internal module and namespace. :wink:

blakeembrey:
It's kind of a PITA, but I'm happy to merge actually. I was thinking `class + namespace` to expose the types. The issue with `export =` is that's the only type you'll be able to access, so you need to work around it (with `namespace`). I'll figure out a PR if we get to there :smile:

unional:
Great. 

Currently I see it working (at least partially?):
image

I run `typings i -D file:gulp.d.ts` and see the result in `test/test.ts`

PR #86, opened 2016-02-04
uuid

rapropos:
Adds typings for [node-uuid](https://github.com/broofa/node-uuid) hosted at [typed-node-uuid](https://github.com/rapropos/typed-node-uuid) under the name "uuid". The project itself is at 1.4.7, but from my understanding of your contribution guide, I set the supported version to 1.4.0. Let me know if this is a misunderstanding.

### DIFF
--- a/npm/uuid.json
+++ b/npm/uuid.json
@@ -1,0 +1,6 @@
+{
+	"versions": {
+		"1.4.0": "github:rapropos/typed-node-uuid#75540df1d026d5c09db1a6b8feba4edf071574d1"
+	}
+}
+
--- a/npm/uuid.json
+++ b/npm/uuid.json
@@ -1,6 +1,6 @@
 {
 	"versions": {
-		"1.4.0": "github:rapropos/typed-node-uuid#75540df1d026d5c09db1a6b8feba4edf071574d1"
+		"1.4.0": "github:rapropos/typed-node-uuid#a2ad36f2802416729eaad89559d76e0b03ba673c"
 	}
 }
 


blakeembrey:
Awesome, looks perfect, cheers! The only thing I noticed was `browser.d.ts` - I think we've got a little misunderstanding, looks like you're trying to type a global interface `uuid`?

rapropos:
My understanding is that the library exposes itself differently depending on whether it is running in a CommonJS environment:

```
  if (('undefined' !== typeof module) && module.exports) {
    // Publish as node.js module
    module.exports = uuid;
  } else if (typeof define === 'function' && define.amd) {
    // Publish as AMD module
    define(function() {return uuid;});
  } else {
    // Publish as global (in browsers)
    _previousRoot = _window.uuid;
    // **`noConflict()` - (browser only) to reset global 'uuid' var**
    uuid.noConflict = function() {
      _window.uuid = _previousRoot;
      return uuid;
    };
    _window.uuid = uuid;
  }
```
...so the various functions are accessed directly from under the import point when in a node or webpack context, but when the script is directly included in a page (as I understand the "browser" context of typings to mean), those functions come under an exported variable `uuid`. I thought this was the proper way of expressing that, but if you have suggestions for a better way, I will incorporate them.

blakeembrey:
Not exactly. The browser story hasn't exactly been solved because it's tricky to type these types of modules (properly). There's a discussion around how to solve it properly (perhaps a different ambient module that exposes these - because browsers are 100% global). For now, it can be done by using TS 1.8+ easily with `declare global` or ambient definitions. However, in this case, the browser field is used for https://github.com/defunctzombie/package-browser-field-spec.

rapropos:
Hmm. When I installed typings, I noticed that it made two parallel sets of .d.ts files, so I thought that those could be provided separately. If you're saying that entries in the registry can only contain a single definition, perhaps everybody would be better served if I just ditched what I was trying to do for the browser case and just try to get what I have in "main.d.ts" incorporated upstream. What do you suggest?

blakeembrey:
No, they can contain both. The browser field is used for following `package.json`-like resolution (see previous link). I noticed you're expected it to be global, but it's not (a feature like that is planned). 

Feel free to delete `browser.d.ts` usage and update the PR, I'm happy to merge it :smile: When proper browser global support is added (either in TS/Typings) I can ping you again.

rapropos:
Do you think "uuid" or "node-uuid" is a better naming choice, or does it not really matter much?

unional:
@rapropos, a "btw suggestion":

You can rebase your branch to the master so that you can keep the PR consist of only 1 commit. :rose:

rapropos:
@unional: thank you for the suggestion. I'm following [this fork synching guide](https://help.github.com/articles/syncing-a-fork/). When I checked out the section on rebasing in [this relatively authoritative-looking source](https://git-scm.com/book/en/v2/Git-Branching-Rebasing), it warns me:

> **Do not rebase commits that exist outside your repository.**
> 
> If you follow that guideline, you’ll be fine. If you don’t, people will hate you, and you’ll be scorned by friends and family.

...which sounds sort of scary.

blakeembrey:
It's ok to rebase, they only exist in your repository still (until I merge).

As for the name, it's fine as is. Feel free to take both if the type definition is compatible with both modules.

rapropos:
I hope I did the rebase properly, please advise if not. If by "both modules", I'm not sure what the other one aside from the one linked in my first comment is. If it's [defunctzombie/node-uuid](https://github.com/defunctzombie/node-uuid), that looks like a fork of the one that I use, but I don't know if any incompatible changes have been introduced.

blakeembrey:
No worries, this is plenty - cheers! :smile:

PR #96, opened 2016-02-08
Typings for jsonwebtoken

enriched:
Typings for [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken) v5.5.4


blakeembrey:
I'm going to merge this, odd that Travis didn't receive the event.

As an FYI, I would recommend refactoring type definitions you copy from DT into the modern format using ES6 `export` instead of `export =` (where it makes sense).

PR #104, opened 2016-02-12
Added typings for npm-toml

Fank:


blakeembrey:
Another FYI, but this typing can be 10x simpler. Just needs to be:

```ts
export function parse(value: string): any;
```

PR #105, opened 2016-02-12
Added mongoose-timestamp

Fank:
mongoose-timestamp require "mongoose" typings which doen't exist currently is this bad?

### DIFF
--- a/npm/mongoose-timestamp.json
+++ b/npm/mongoose-timestamp.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.5.0": "github:Fankserver/npm-mongoose-timestamp#67e521f1a25a2a21f25f7b74a7227ef36e0a8a7c"
+  }
+}


blakeembrey:
Nope, it's ok. Mongoose is more of a "peer dependency" in this case anyway, but we should type Mongoose ASAP :+1:

Fank:
Ill try to do this soon

blakeembrey:
No rush. Thank you for your help :smile:

stevehipwell:
Is there any progress on the mongoose typings?

Fank:
not now but it is in my personal pipeline (~3 weeks)

stevehipwell:
@Fank thanks for the quick reply. I've added the current definitely typed mongoos definition into my repo and removed the promise class definition that conflicts with the es6 promise declaration as a workaround so I can use catch.

Any news on the global node definition?

blakeembrey:
> Any news on the global node definition?

This one is safe to use from DefinitelyTyped. Eventually it can be converted, but the usefulness of that is minimal since it's one of the only definitions actually correct according to how Typings works.

PR #106, opened 2016-02-12
Added gravatar

Fank:


blakeembrey:
An FYI, but this could easier written I believe. It'd also be great to have options typed:

```ts
declare function gravatar (email: string, options?: gravatar.Options, addProtocol?: boolean): string;

declare namespace gravatar {
  export interface Options {
    s?: number | string;
    size?: number | string;
    r?: string;
    rating?: string;
    d?: string;
    default?: string;
    f?: string;
    forcedefault?: string;
  }
}

export = gravatar;
```

I'll open a PR and you can update this again.

PR #108, opened 2016-02-12
Updated gravatar

Fank:
The url function were lost


blakeembrey:
Ah, you're right! I completely missed that somehow, sorry! In that case, you can kill the namespace:

```
export function url (email: string, options?: Options, addProtocol?: boolean): string;

export interface Options {
  s?: number | string;
  size?: number | string;
  r?: string;
  rating?: string;
  d?: string;
  default?: string;
  f?: string;
  forcedefault?: string;
}
```

PR #112, opened 2016-02-13
Added node-teamspeak typings

Fank:
How do i add (peer)dependencies for node https://github.com/Fankserver/npm-node-teamspeak/blob/fd84297b37aa18517360a8d3c3af5bda8a1a5a25/node-teamspeak.d.ts#L1 ?


blakeembrey:
In this case, it's not really a peer dependency. Ambient dependency usage is correct. You can always kill the extra module wrapper you write too and just export the class.

PR #113, opened 2016-02-13
Added npm simple-xmpp typing

Fank:
https://github.com/Fankserver/npm-simple-xmpp

### DIFF
--- a/npm/simple-xmpp.json
+++ b/npm/simple-xmpp.json
@@ -1,0 +1,5 @@
+{
+	"versions": {
+		"1.3.0": "github:Fankserver/npm-simple-xmpp#8ce6042ecd0538cd3a352fe881b10e16384fa31d"
+	}
+}


blakeembrey:
Awesome :+1: Would be interesting to see this typed with the 1.8 features like string literals for the events (just a thought).

Fank:
I'm not that familiar with typings (currently its more like try and guess :D) could you may provide an example what you mean with "string literals"?

blakeembrey:
@Fank Honestly, it was just an idea and it's a TypeScript feature. I'm just to use it, but I know I can use it for features like this.

References:

* https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#string-literal-types
* https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L4284-L4292

Fank:
:smile: i've searched for something like this. i tried to do it with an "enum" but it didn't worked. This is awesome.
I added it in #122

PR #115, opened 2016-02-14
react-router

andrew-w-ross:
Updated the history definition.
Added definitions for react-router.


andrew-w-ross:
Ah dang, I thought I did a rebase. Let me attempt to fix it.

PR #116, opened 2016-02-14
Updates and additions

andrew-w-ross:
Updated the history type definitions to use interfaces instead of types.
Added definitions for react-router-redux.
Added the react-router definitions.
Updated the redux definitions to be s5 compatible.


blakeembrey:
Thanks :+1:

PR #117, opened 2016-02-15
Added voximplant-websdk

aylarov:


Fank:
Remove this https://github.com/aylarov/TypeDefinitions/blob/master/voximplant-websdk/voximplant-websdk.d.ts#L2039-L2041 it will automatically added on typings compile

blakeembrey:
@aylarov Are you aware there's no export from your definition? Since this is an external module, you probably meant to export something. One solution would be to add `export = VoxImplant;` to the bottom, or move the proper exports out of the namespace and export them normally. Typings will automatically wrap your module in a `declare module "..."` declaration.

PR #129, opened 2016-02-18
Added webpack

tkqubo:
This webpack type definition https://github.com/tkqubo/typed-webpack is almost simply copy-and-pasted from my original definitions committed to DefenitelyTyped repo.  Now I want to maintain this definition within typings ecosystem.  But I'm still not sure `npm/webpack.json` is the right place.  Anyway I'd appreciate if maintainer would check my PR :bow:

### DIFF
--- a/npm/webpack.json
+++ b/npm/webpack.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.12.9": "github:tkqubo/typed-webpack#49876a3de79056dd978ea8a68f0cf17d7332826f"
+    "1.12.9": "github:tkqubo/typed-webpack#642ef998dda5b1da61206122ad3a302b5845542a"
   }
 }
--- a/npm/webpack.json
+++ b/npm/webpack.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.12.9": "github:tkqubo/typed-webpack#642ef998dda5b1da61206122ad3a302b5845542a"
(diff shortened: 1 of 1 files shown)


Fank:
Hey just copy pasting this won't work. You added an ambient `declare module "MODULE"` https://github.com/tkqubo/typed-webpack/blob/master/webpack.d.ts#L7

Which isn't needed because it will automatically added on typescript compile.
Here is an working example which may helps you to understand how a definition should looks like https://github.com/Fankserver/npm-node-teamspeak/blob/master/node-teamspeak.d.ts

tkqubo:
@Fank Thanks so much, your advice helped me a lot!  Sorry for my poor understanding of the concepts.

Fank:
@tkqubo I opened #130, if you think there some information which should be added than add a comment

tkqubo:
@Fank  OK I see.  I will write some later if any :bow:

blakeembrey:
This looks great. What would it take to type `uglify` so it's not ambient? I'd prefer not to merge dependencies that don't work out of the box.

A workaround for having to type `uglify` would be to move the uglify options inline so there's no need for the import anymore. What do you think?

tkqubo:
@blakeembrey  You are right.  So far I don't know to what extent the dependencies should be apparent on the type definitions in line with the original JavaScript module . But since `uglify`'s only usage is `UglifyJS.MinifyOptions`, inlining this can be very reasonable.  Thanks!

blakeembrey:
Amazing! :+1:

tkqubo:
Thanks for you guys! @blakeembrey @Fank

PR #131, opened 2016-02-19
Add npm-assertion-error

unional:


blakeembrey:
There's also `toJSON` :+1: https://github.com/chaijs/assertion-error/blob/master/index.js#L106

unional:
Man, how do you spot that! :+1: Will add

PR #133, opened 2016-02-21
Add react-highlighter.

starhoshi:
https://github.com/starhoshi/typings-react-highlighter


starhoshi:
Sorry, pull request again later.

blakeembrey:
@starhoshi It's failing because you're using an ambient declaration. If you switch to non-ambient it'll work (just move everything inside of `declare module "..."` outside).

PR #134, opened 2016-02-22
Add chai

unional:
Port from DT.

Some signatures is not complete (e.g. new chai.Assertion()).  Will add in the future.

### DIFF
--- a/npm/chai.json
+++ b/npm/chai.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.4.0": "github:typed-typings/npm-chai#f74b0d35e090039f303f6fef184c6f102408a9fc"
+    "3.4.0": "github:typed-typings/npm-chai#450bf667f66ae45b23e72bd836d2b44835413df0"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
Nothing is expected in the end? https://github.com/typed-typings/npm-chai/blob/master/lib/index.d.ts

unional:
Not exactly a typo. I tried `typings i file:lib/index.d.ts` and it gives me the right type in `test/test.ts`. So I try to keep it without. Is this a bug or a side effect?

I can put `export = chai;` in at the end.

blakeembrey:
That's odd, seems like it should break.

Also, what's the intention of https://github.com/typed-typings/npm-chai/blob/master/chai.global.d.ts? Looks like the import is wrong, but should it transform as part of `index.d.ts`?

unional:
Yes the `chai.global.d.ts` should goes inside of `index.d.ts`. Updating.  By the way, how can I add the property to object literal or `any`? 
The original code in DT is written as:
```
interface Object {
   should: Chai.Assertion
}
```

It works if `var x: Object` but not if it is object literal `var x = {};`. i.e.:

```js
var x = new Object();
x.should... // works
var y = {};
y.should... // does not work
```

unional:
@blakeembrey , anything you want to change before merging this?

blakeembrey:
I haven't had a chance to test just yet, but if you're comfortable with it let's merge :+1:

PR #136, opened 2016-02-22
sequelize typings

louy:


blakeembrey:
A lot of this project won't work unless the user knows to install all these dependencies. Is it possible to inline or re-create the dependencies before we merge this? At least with DefinitelyTyped it'll still warn on the stripped references to let the user know about the missing dependencies.

louy:
so the only ambient dependency now is lodash. I wonder why didn't anyone move it to typings yet.

blakeembrey:
Did you want to port it?

louy:
**lodash** is a huge project and I really don't want to maintain its typings because I don't use it.

@bczengel @chrootsu would you guys be kind enough to migrate lodash to typings? I see you're still actively maintaining it on DefinitelyTyped.

blakeembrey:
We can certainly start a repo under https://github.com/typed-typings too, where we can add whoever wishes to help with maintenance.

unional:
Done! :smirk:  Currently it is completely empty

bczengel:
Is there a guide for porting from DT to typings?  I did some googling and found a bunch of details but I'm having trouble pulling it all together.

blakeembrey:
@bczengel There's not a lot involved, and there's a lot of repos in https://github.com/typed-typings. The only main difference is using `typings.json` and using external modules instead of ambient definitions.

blakeembrey:
@louy Let me know when/if you're happy with this to be merged. I think if people, not you, run into issues, we can pull you in or point them to your repo anyway.

alitaheri:
Any movement on this one? Lodash is already in the organization. should we move this one there too? @louy

louy:
The only thing left is #334. I'll update this PR as soon as that one is merged 👍

alitaheri:
@louy Awesome job as always :heart_eyes: Thanks :+1: :+1:

louy:
@blakeembrey This one is ready to be merged.
@unional thank you!

PR #137, opened 2016-02-23
Add aurelia-pal

unional:
Blake, aurelia-pal is on beta and its version is 1.0.0-beta.1.1.1

https://github.com/aurelia/pal/releases

Anything I need to be concern of?

I set the version to 1.0.0

### DIFF
--- a/npm/aurelia-pal.json
+++ b/npm/aurelia-pal.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:typed-typings/npm-aurelia-pal#c14d54ee9f1b6deb442f3d2499ccf22c6886fe2f"
+    "1.0.0-beta.1.1.1": "github:typed-typings/npm-aurelia-pal#c14d54ee9f1b6deb442f3d2499ccf22c6886fe2f"
   }
 }
--- a/npm/aurelia-pal.json
+++ b/npm/aurelia-pal.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:typed-typings/npm-aurelia-pal#c14d54ee9f1b6deb442f3d2499ccf22c6886fe2f"
+  }
+}


blakeembrey:
If you want, I suppose you could use the beta version as the key? I assume because it's beta things might change before the final release.

unional:
Ok. Just did.  Didn't do that in the beginning because I thought it must comply to semver.

unional:
> [Error: Invalid JSON for "aurelia-pal (npm)":
data.versions should NOT have additional properties]

`1.0.0-beta.1.1.1` fails "^\\d+\\.\\d+\\.\\d+(?:-\\w+)?$" in schema json.

Need to expand it to:
"^\\d+\\.\\d+\\.\\d+(?:-\\w+)?\\.\\d+\\.\\d+\\.\\d+$/"?

I can create a PR for that.

blakeembrey:
No, just add `\.` to the matching group. I can update it. Pre-release versions are part of semver.

blakeembrey:
Ok, pushed the semver correction in better accordance with the spec (still not 100%, but 99% there).

unional:
:+1:

blakeembrey:
That's interesting. LGTM. I actually thought I was reading through TypeScript looking at the source code, but I guess that's Flow? I wonder if TypeScript could parse that file and generate a `.d.ts` file. Might be worth support `.ts` files in Typings in the future.

unional:
Aurelia is a different beast.  They are written in JS with type and the build script creates the `.d.ts` by directly managing the imports using a util.  No pass to tsc at all.

unional:
I can't really tell it is not TypeScript either. :smile:

unional:
I don't think it is Flow thou.

blakeembrey:
If they generate `.d.ts` files, is there a need to re-type them?

unional:
The `.d.ts` they have is ambient. Also npm distribute only the `.js` and `.d.ts`.

blakeembrey:
Oh, ok. Have you submitted an issue to them about the ambient definitions? Seems like a useful fix so TypeScript users can actually use their definitions.

unional:
Yes I can. But do we want to settle the NPM registry / 1.0 roadmap first? Seems like that could change how the source author add typings to the repo (i.e. package.json `typings` property vs `peerDependencies`).

blakeembrey:
No, that won't change anything for the source author. That discussion is only for consume type definitions of modules that are distributed without `.d.ts` files.

blakeembrey:
I'll merge it for now, we can always remove it again when Aurelia corrects their definitions.

unional:
Ok. Moving discussion to gitter. :+1:

PR #138, opened 2016-02-23
Added typings for datautils package at v3.1.4

riggerthegeek:


blakeembrey:
@riggerthegeek You don't actually need to publish here for your typings to work with TypeScript. Just add a `"typings": "datautils.d.ts"` field to your `package.json` and publish away.

Just a note, using https://github.com/riggerthegeek/datautils-js/blob/master/package.json#L6 with a non-ambient definition will create issues too. I'm not aware of any other tools using it, so I assume that's for TSD, but TSD expects only ambient definitions.

blakeembrey:
Here's an example of a JavaScript package I've written with the `typings` field in `package.json` for TypeScript users: https://github.com/blakeembrey/change-case/blob/master/package.json#L6

riggerthegeek:
@blakeembrey thanks Blake, I'll do that. I'd forgotten that the `typings` field was in there.

I didn't know about the `typings` field in `package.json`.  Might I suggest sticking that in the Readme page for the project?

blakeembrey:
Sure, sounds like a good idea :+1: In https://github.com/typings/typings? 

There's https://github.com/typings/typings#should-i-use-the-typings-field-in-packagejson, but it could definitely be clearer. The `typings` field is explicitly a TypeScript feature, whereas Typings (the tool) is made to sort of "polyfill" non-typed modules.

riggerthegeek:
I'd stick it in both https://github.com/typings/typings and https://github.com/typings/registry.  It doesn't need a lot - if you copied your second response to me verbatim that would probably do the job.

I didn't find the "should i use the typings field" overly clear.  Perhaps a second paragraph of two sentences - first explaining a good use-case for using the `typings` field and the second being a good use-case against.

Thanks for your help

PR #142, opened 2016-02-25
added ssv-core library

stephenlautier:


blakeembrey:
@stephenlautier If it's a regular package, you don't actually need to add it to the Typings registry. I answered this recently too: https://github.com/typings/registry/pull/138.

stephenlautier:
Ah oki will do that, thought it make sense to be installable by a shortname if needed.

Does the package.json get resolved from anywhere within the project? I'm mainly using JSPM so it would look like 

```
root
 |---- wwwroot/jspm_packages/npm/sketch7/ssv-core/package.json
```

I'm asking because at the moment i dont even have the package.json within the library when installed

stephenlautier:
I had actually test it out quickly by creating a quick package.json.
Sorry if this is a dumb question, but then how would I install it? via `typings install`

blakeembrey:
Is there a `typings.json` file? You can use `typings install file:<path>`.

blakeembrey:
Sorry, I'm not really familiar with JSPM here. @unional?

stephenlautier:
JSPM works similarly to bower or npm pretty much, from files point of view, just a little bit different structure where you have `jspm_packages`. The `jspm_packages` can be placed not only in the root of the project, you can specify it within `packages.json` via:

```json
"jspm": {
    "directories": {
      "baseURL": "wwwroot"
    },
}
```

As a dev I expect to install typings either via cmd `typings install xxx`, directly from the repo e.g.
```
typings install ssv-core 
```
As thats how package managers generally work, at least npm, jspm and bower

or else if the library I downloaded from bower/jspm/npm or so contains the definitions with it, maybe something like what `tsd link` used to do "might" be good. But personally I would prefer install it directly via `typings install xxx`, as `tsd link` used to install a lot of stuff from `node_modules` dependencies which i don't care about.

Hence that's why I tried to provide typings on your registry, cause currently I was using:
```
typings install github:sketch7/ssv-core --save

// but i wanted it to be more sexy as below
typings install ssv-core --save
```

unional:
@stephenlautier I'm with you on `jspm` support. :smile:

As @blakeembrey said, you should put the typings directly in your package. `jspm` support is a feature in discussion: https://github.com/typings/typings/issues/238

Typings registry is not for the shorthands as in jspm registry.

I know where you are coming from. We get pretty spoiled by the `jspm install ssv-core` syntax. :rose:

stephenlautier:
@unional opps i thought it was the same idea with the typings registry :D. I will follow the jspm support. Thanks!

PR #143, opened 2016-02-25
feat: add swagger-express-middleware type definition

tkqubo:


blakeembrey:
LGTM, though I don't usually like coupling of modules like that - the definition can break at any time if Express changes.

tkqubo:
Thanks for merging!  But sorry for adding dependencies without thinking too much.  I'll fix it later. :bow:

PR #145, opened 2016-02-26
added reference to zoomcharts.d.ts

Knagis:
The readme states that the typings have to be pushed to github repository. Is this mandatory? We already publish the .d.ts file with every release to our CDN that could be linked directly.

Would a reference like in this PR this be allowed in the registry?

Also I would like to better understand the value of specifying URLs to different versions. We have a definitions file attached to each version of the library - should I add those explicitly, like "1.11.0", "1.11.1", etc.? In the PR the last entry actually points to the CDN link that redirects to the most latest - which isn't exactly great considering that after few months the definition file might not match the 1.12 version anymore...

### DIFF
--- a/global/zoomcharts.json
+++ b/global/zoomcharts.json
@@ -1,0 +1,8 @@
+{
+  "versions": {
+    "1.9.0": "https://cdn.zoomcharts-cloud.com/1/9/latest/zoomcharts.d.ts",
+    "1.10.0": "https://cdn.zoomcharts-cloud.com/1/10/latest/zoomcharts.d.ts",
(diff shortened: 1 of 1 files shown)


blakeembrey:
They should just be the minimum semver-compatible releases. Also, any Typings-compatible scheme is available, so yes - URLs are awesome too :+1: It might be a little better to be completely explicit, but I'm happy to merge this for now. Preferably entries in the registry are always immutable (since these are going to be used by clients directly, so new changes will break them).

Knagis:
OK, understood. I will probably try to incorporate a semi-automatic PR creation in our release process.

blakeembrey:
Let me know if I can help at all. I also want to build the same process into the tooling, so if there's a way to make that work for you we can collaborate.

PR #160, opened 2016-03-01
Fix empty module export.

unional:
@blakeembrey, when I run `ts-node test/test.ts`, I got this error:
```sh
⨯ Unable to compile TypeScript

bundle.d.ts (460,5): Duplicate identifier 'should'. (2300)
lib/Chai.d.ts (29,5): Duplicate identifier 'should'. (2300)
```

The typings seems to be generated correctly (I tried it with `typings i file:typings.sjon`).

Seems like ts-node will compile based on `tsconfig.json`, which has reference to `index.d.ts` (thus loading `lib/Chai.d.ts`), and then run `test/test.ts` (thus loading `bundle.d.ts`).

Also, the seems to be an issue on the global augmentation. If two modules both augment global with the same name (`Object.should` in this case), the compilation fails?

### DIFF
--- a/npm/chai.json
+++ b/npm/chai.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.4.0": "github:typed-typings/npm-chai#450bf667f66ae45b23e72bd836d2b44835413df0"
+    "3.4.0": "github:typed-typings/npm-chai#01a92b8efc0cfa0ac894d695abd06b8b11e3b75b"
   }
 }


unional:
The `global` part is still not working... when I do `typings i github:typed-typings/npm-chai`.

unional:
Seems like the `declare` in `declare global` is stripped out:

```ts
// Compiled using typings@0.6.8
// Source: lib/Chai.d.ts
declare module 'c/lib/Chai' {
import {Assertion} from 'c/lib/Assertion';
import {AssertionError} from 'c~assertion-error';
import {AssertStatic} from 'c/lib/Assert';
import {AssertionStatic} from 'c/lib/Assertion';
import {ExpectStatic} from 'c/lib/Expect';
import {Should} from 'c/lib/Should';
import {Config} from 'c/lib/Config';
import {utils} from 'c/lib/utils';

export interface Chai {
  AssertionError: typeof AssertionError;
  Assertion: AssertionStatic;
  expect: ExpectStatic;
  assert: AssertStatic;
  config: Config;
  util: utils;
  should(): Should;

  /**
   * Provides a way to extend the internals of Chai
   */
  use(fn: (chai: any, utils: utils) => void): Chai;
}

global {
  interface Object {
    should: Assertion;
  }
}
}
```

blakeembrey:
@unional Great catch! It is stripped out, gah! It's a Typings issue, since it's stripping `declare` when putting in inside the `declare module "..."`.

blakeembrey:
@unional Actually, just tested it, it's correct. I need to strip `declare` (including from `global`) to make it work in the ambient context. Just did a test to confirm, but it appears to work without `declare` once transformed.

unional:
That explains why the duplication error `tsc` throws. Do you see the `Object` augmented with `should`? I tried it and it is not appearing.

blakeembrey:
I did in my own tests, just cloning the repo now to check what's going on. When I was using `typings`, the output was completely different - but now I realize I'm probably using a pre-1.8 TypeScript Typings version.

blakeembrey:
@unional Fixed the tests with https://github.com/typed-typings/npm-chai/commit/46fdfe69cc0488bad30121743f796b9fa9de2370. Happy for you to merge when you're ready :+1:

unional:
@blakeembrey , the `global` is still not working when I load it from a different project. 
I created a test project to demo it:
https://github.com/unional/test-npm-chai

blakeembrey:
@unional Works for me. How are you testing it? Remember to exclude `browser.d.ts` (sorry!). Also, if you're relying on an IDE, remember some aren't yet on TypeScript 1.8 (only recently released). Also, make sure you installed with Typings 0.6.8 - earlier versions are using an earlier TypeScript.

```
     "exclude": [
-        "node_modules"
+        "node_modules",
+        "typings/browser",
+        "typings/browser.d.ts"
     ]
```

unional:
It doesn't work for me on both IDE and command line:
image
```sh
MacBook-Pro:test-npm-chai hwong$ ts-node -n test.ts

/Users/hwong/github/unional/test-npm-chai/test.ts:5
someObject.should.be.an('Object');
          ^
TypeError: Cannot read property 'should' of undefined
    at Object.<anonymous> (/Users/hwong/github/unional/test-npm-chai/test.ts:5:11)
```

I am using TypeScript `1.8.2` But I guess it would make things better, not worse? :smirk:

Just test with `1.7.3`, same error thou. :cry:

blakeembrey:
@unional Oh, yeah, that's why I pointed to the commit. Chai requires that you actually "install" it, so the baseline is actually wrong. See https://github.com/typed-typings/npm-chai/commit/46fdfe69cc0488bad30121743f796b9fa9de2370#diff-632988bf801a8c538f18e7c4e222d9caR4.

blakeembrey:
However, your IDE won't work with this until it uses TypeScript 1.8 - Atom is already updated, but I think VS Code might take a few days.

unional:
Arrrrh~~~~
Interesting.  I can see it working now. How does `tsc` tell when to apply the augmentation? The typings does not specify the relationship between the augmentation and `should()`:
```ts
export interface Chai {
  AssertionError: typeof AssertionError;
  Assertion: AssertionStatic;
  expect: ExpectStatic;
  assert: AssertStatic;
  config: Config;
  util: utils;
  should(): Should;

  /**
   * Provides a way to extend the internals of Chai
   */
  use(fn: (chai: any, utils: utils) => void): Chai;
}

declare global {
  interface Object {
    should: Assertion;
  }
}
```

btw, you can configure vscode to use a local copy of TypeScript by:
```js
// /.vscode/settings.json
{
  "typescript.tsdk": "node_modules/typescript/lib"
}
```

unional:
.....something is weird. The IDE is working now with TS@1.8.2.  It wasn't before. Can't reproduce it.

Yes, noticed the `should()` makes the test pass. Completely forgot about calling `should()` because I primary use `karma-chai` which makes the call automatically. :rose:

blakeembrey:
Oh, fair enough - I'd need it per-project though. Anyway, it's not possible, sadly. This should be a feature request of TypeScript, but I think it's actually a really hard thing to solve - since you're basically typing side-effects. If TypeScript could make it happen though, that'd be huge!

unional:
Do you mean vscode? Yes, it is per project.

blakeembrey:
VSCode, yes, cool - I'll try to use that in the future. I wonder if they support older versions too.

unional:
Don't know how far it goes backward. Just tried 1.7.3. :smile:

unional:
Thank!

PR #161, opened 2016-03-01
griddle-react definitions

hodavidhara:
Definition for [Griddle](http://griddlegriddle.github.io/Griddle/).

They aren't yet fully complete to my liking, but they're much better than just an any definition.


blakeembrey:
Thanks! :+1:

PR #163, opened 2016-03-01
object-assign typings

louy:
https://www.npmjs.com/package/object-assign


blakeembrey:
LGTM. Looks like Travis never got the memo here.

PR #168, opened 2016-03-03
Added retext definitions

joewashear007:
* Added the definitions file for retext and its dependencies


wooorm:
Thanks for the ping, looks like wonderful work. Unfortunately I don't know anything about these typing and I'm not using them, making me not very good at potentially maintaining them!

blakeembrey:
No worries :smile: If the public API changes, feel free to ping me or @joewashear007 and we can keep the TypeScript type definitions in sync.

wooorm:
:+1: I’ll try my best!

PR #153, opened 2016-02-28
Adding typings for primus.

dercodebearer:
Resolves issue #152

PR #176, opened 2016-03-05
Add typings for BottleJS

kenvunz:


blakeembrey:
Same comment as https://github.com/typings/registry/pull/153#issuecomment-189942337.

blakeembrey:
LGTM, I think there's some aliases with `export =` happening but you can let us know if there's an issue and push an update. Thanks for contributing :smile:

PR #182, opened 2016-03-07
npm-through2

### LINKED ISSUE #180: through2

unional:
closes #180

### DIFF
--- a/npm/through2.json
+++ b/npm/through2.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.0.0": "github:typed-typings/npm-through2#772fb54249feb63e217010bcc6ab5ff72e2d314a"
+    "2.0.0": "github:typed-typings/npm-through2#2b1d39490ad8530546d97babee8cfaaabed9e0cd"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
You've got a couple of typos in the definition: `TransfofmCallback` and `FlashCallback`. Also, `push` (https://github.com/typed-typings/npm-through2/blob/master/index.d.ts#L19) does not exist on `through2` - it's a part of `this` in the callback function. I can't recall if you can type `this` in TypeScript yet, but at least it could just be an interface of what `this` should be so you can use it.

unional:
> I can't recall if you can type this in TypeScript yet, but at least it could just be an interface of what this should be so you can use it.

So comment it out for now?

```ts
  // push() is not on through2, it's on `this` in the callback.
  // Don't have a way to type this yet.
  // export function push(data: any): void;
```

blakeembrey:
You can also provide it typed as an interface so people can use it, if needed `export interface This { push() }`.

blakeembrey:
Feel free to merge when you're ready.

PR #192, opened 2016-03-08
Update for typings 0.7 and some cleanup

unional:
Improve namings

### DIFF
--- a/npm/chai.json
+++ b/npm/chai.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.4.0": "github:typed-typings/npm-chai#07a92689f3de9a137f75267f02f1097a9a351a5c"
+    "3.4.0": "github:typed-typings/npm-chai#e8db615cea6be222a9547bead73ec66b82691108"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
Looks like a bug in Typings programmatic API, I'll take a look but I'd guess I have some updates to do.

unional:
Could it be because my `npm test` didn't include `typings bundle`? I have it separated.

unional:
`npm run build` = `typings bundle -out test`
`npm test` = ts-node -n test/test.ts`

blakeembrey:
No, it's because the API changed between 0.6 and 0.7 and I just merged 0.7 into master.

blakeembrey:
@unional The last issue is case-sensitivity. Looks like you're using `utils.d.ts` (lower-case), but it's referenced using upper-case (`Utils`).

unional:
Yeah, I updated it locally (on osx). Seems like git push didn't push it up.

blakeembrey:
Renaming case-sensitivity with Git is a PITA, it's ok :+1:

blakeembrey:
Thankfully, you were the only one to use the `.d.ts` paths so no one else needs changing right now :smile: I'll get around to doing properly file resolution when I can.

unional:
I thought that problem only exists on perforce. ....:P

unional:
> Thankfully, you were the only one to use the .d.ts paths so no one else needs changing right now :smile: I'll get around to doing properly file resolution when I can.

Figuring out the best practice along with `jspm`/`systemjs`. :(

PR #197, opened 2016-03-09
added reference for jasmine-node-promise-matchers

adamball:
Typings URL: https://github.com/adamball/npm-jasmine-node-promise-matchers


blakeembrey:
Unfortunately this won't work for people using non-DT Jasmine typings, but that should be fine - we'll ping you if someone makes an issue about it. Thanks for contributing!

PR #199, opened 2016-03-09
fix for jasmine-node-promise-matchers

adamball:
Typings URL: https://github.com/adamball/typed-jasmine-node-promise-matchers

previous type definition was causing compile time errors, this PR contains the fix for that.

### DIFF
--- a/npm/jasmine-node-promise-matchers.json
+++ b/npm/jasmine-node-promise-matchers.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.0.3": "github:adamball/npm-jasmine-node-promise-matchers#56c07a89cf9864bc22501449a6d05af2b2ce4e1e"
+    "0.0.3": "github:adamball/typed-jasmine-node-promise-matchers#139a84fa30e3edfa40197a16a50ec5c029991a07"
   }
 }


adamball:
I've realized that the definitions that I wrote need to be ambient (and I did just that in this PR).  As per the FAQ section "How Do I Write Typings Definitions?":

https://github.com/typings/typings/blob/master/docs/faq.md

I set the ambient flag to true in my typings.json file:

https://raw.githubusercontent.com/adamball/typed-jasmine-node-promise-matchers/master/typings.json

but it is showing as non-ambient in the typings CLI search and install.  

How do I go about properly configuring this json file in the registry so that it searches and installs as ambient?  If this isn't possible, should I post this to DT instead?

Thanks for the help as I'm kind of new to all of this.

blakeembrey:
> How do I go about properly configuring this json file in the registry so that it searches and installs as ambient? If this isn't possible, should I post this to DT instead?

I'm not entirely sure what you're asking, but I'll try to clarify what it's doing. When you install normally, it's prefixing `npm!<pkg>`, when you search ambient it's prefixing `dt!<pkg>`. Those defaults can be changed at any time. 

For search, that sounds like a bug to me. I'll make a patch so it always shows up. I also have another PR coming to error when you've set ambient to true and it's installed to non-ambient. Does that help answer what's happening?

blakeembrey:
Also, NPM is non-ambient by default (https://github.com/typings/registry#structure). You can move it to `global` if you wanted it to be "ambient" (though that distinction matters less now than in 0.6).

adamball:
That answers my question, thank you.  I seem to have missed the 'Structure' section in my reading of the README.md before.

I'll work on another PR to move the json file to global as per your suggestion.  I believe that's the proper place for it.  Thank you for your detailed responses.

PR #200, opened 2016-03-09
Added typings for codemirror-meta addon

bossqone:
Typings URL: [https://github.com/bossqone/typings-codemirror-meta]
Source URL: [https://github.com/codemirror/CodeMirror/blob/master/mode/meta.js]


blakeembrey:
Thanks!

PR #201, opened 2016-03-09
Added typings for codemirror-loadmode addon

bossqone:
Typings URL: [https://github.com/bossqone/typings-codemirror-loadmode]
Source URL: [https://github.com/codemirror/CodeMirror/blob/master/addon/mode/loadmode.js]

### DIFF
--- a/global/codemirror-loadmode.json
+++ b/global/codemirror-loadmode.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "5.11.0": "github:bossqone/typings-codemirror-loadmode#1bb1da929f2405c55a67c7d2672639f241c9607e"
+  }
+}


blakeembrey:
Cheers! It'd be cool to get some of these to use module augmentation with 1.8 so it'll work with future CoreMirror type definitions, but it's good to merge.

bossqone:
May I ask a few questions? This was my first TypeScript contribution, so I'm a little confused.
I've installed this module using: `node_modules\.bin\typings install global!codemirror-loadmode --save`

Here's the downloaded definition:
```typescript
// Generated by typings
// Source: https://raw.githubusercontent.com/bossqone/typings-codemirror-loadmode/1bb1da929f2405c55a67c7d2672639f241c9607e/codemirror-loadmode.d.ts
declare module '~codemirror-loadmode/codemirror-loadmode' {

module CodeMirror {
    var modeURL: string;
    
    function requireMode(mode: string, cont: () => void): void;
    function requireMode(mode: { name: string }, cont: () => void): void;
    
    function autoLoadMode(instance: CodeMirror.Editor, mode: string): void;
    function autoLoadMode(instance: CodeMirror.Editor, mode: { name: string }): void;
}
}
```

1.) Is this ok ? Why is it wrapped ?
2.) Shouldn't typings manager download the dependencies as well? The module contains ambientDependencies entry in typings.json, but the `typings` directory doesn't contain this dependency.

Thanks.

blakeembrey:
1) You didn't use `--ambient` to install. If you want to enforce ambient, you can should set a field called ambient in `typings.json` to `true` and I'll make a fix right now to error during install on that.
2) No, `typings` does not resolve ambient dependencies. You might be interested in https://github.com/typings/typings/issues/281#issuecomment-192551121.

blakeembrey:
The difference between ambient is pretty well summed up in a few places (too many places, I should write it in the FAQ for future questions). Basically:

* We can't version with ambient modules
* We can't namespace with ambient modules
* And because of these issues, using ambient modules will break your consumers code if we bundle them with your package - so we don't

blakeembrey:
Does that kind of help? Hopefully enabling the flag in `typings.json` will improve this experience - I've been trying to auto-detect whether something looks ambient in the past and error but it fails here because there's no real indication of which it is.

bossqone:
Thanks, `--ambient` solved my problem for now.

CodeMirror definition already exists in the DefinitelyTyped repo. If I understand, owners of that typing definition (in DT) should port it to the `typings/registry` repo. Otherwise, there'll be problems (or inconsistency) with `ambient`.

So what are my possibilities? How can I help to improve current state.
Please correct me if I'm wrong.

blakeembrey:
Technically, yes. But it should be ok. We can improve things by trying to collaborate with authors in DefinitelyTyped or across the ecosystem more, been meaning to create an issue about this.

On a side note, I'm really glad doing `global!` clicked. I was afraid documentation was too sparse since it's new in 0.7 - how did you find you needed to do that?

bossqone:
I have noticed that `typings install codemirror-loadmode` tried to install from `/npm/*`, so I needed to specify another source. `typings install --help` was really helpful.

PR #202, opened 2016-03-09
fixed EventEmmiter returning this. ref goenning/typed-pg#1

goenning:
This is just a small fix for some methods returning wrong type.

Typings URL: https://github.com/goenning/typed-pg
Source URL: https://github.com/brianc/node-postgres


blakeembrey:
Thanks!

PR #205, opened 2016-03-10
jasmine-node-promise-matchers moved to global directory

adamball:
Changes as per discussion in:

https://github.com/typings/registry/pull/199

Typings URL: https://github.com/adamball/typed-jasmine-node-promise-matchers


adamball:
no problem at all, i've updated my local typings.json to reference the direct github link and it works just fine.

PR #208, opened 2016-03-15
Add typings for password-generator module

SomaticIT:
Typings URL: [https://github.com/typed-contrib/password-generator]
Source URL: [https://github.com/bermi/password-generator]


blakeembrey:
Only need `node.d.ts` again.

SomaticIT:
Ok I updated the project and the registry to remove unnecessary browser typings and since typings are valid from range 0.1.0 to 2.0.2 I changed registry min version to 0.1.0. Is it correct ?

blakeembrey:
Seems perfect, cheers! It'd be good to get proper version ranges here, but I'll update you if that happens.

PR #211, opened 2016-03-15
Add typings for gaikan module

SomaticIT:
Typings URL: [https://github.com/typed-contrib/gaikan]
Source URL: [https://github.com/Deathspike/gaikan]

### DIFF
--- a/npm/gaikan.json
+++ b/npm/gaikan.json
@@ -1,5 +1,5 @@
 { 
     "versions": { 
-        "2.0.2": "github:typed-contrib/gaikan#59199e796186563187305a80fac20ea7ca06e8cb" 
+        "2.0.0": "github:typed-contrib/gaikan#8c1f686cea586380f04cdac4284a65f5bb2b3d70" 
     }
 } 
--- a/npm/gaikan.json
+++ b/npm/gaikan.json
@@ -1,5 +1,5 @@
 { 
     "versions": { 
-        "2.0.2 ": "github:typed-contrib/gaikan#59199e796186563187305a80fac20ea7ca06e8cb" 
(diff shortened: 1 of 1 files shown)


blakeembrey:
I think you'll need to remove `browser` from your typings, not sure why it's there. Is there a browser resolution version on the package?

SomaticIT:
The gaikan module is different depdending on the context.
In node, it contains more properties and more functions.
Maybe should I do like I did with knockout (#212 #213) ?

blakeembrey:
Probably, yes. Though the TypeScript UMD support would be better to hold out for. Unfortunately, `browser` does not do what you think it does. See https://github.com/typings/typings/blob/master/docs/faq.md#how-do-i-write-typings-definitions and https://github.com/defunctzombie/package-browser-field-spec.

SomaticIT:
@blakeembrey I think I understood correctly the browser entry in package.json.

As you can see in gaikan documentation, there is two different API depending on the context :
https://github.com/Deathspike/gaikan#7-server-api
https://github.com/Deathspike/gaikan#8-browser-api

Moreover if you look at the entry point file [index.js](https://github.com/Deathspike/gaikan/blob/master/lib/index.js), you can notice that the real entry point is different depending on the context :
```javascript
module.exports = typeof window === 'undefined' ? 
        require('./server') : 
        require('./browser'); 
```

So I think my typings are correctly defined. I improved them during my tests.

blakeembrey:
Ah, you're right. This is kind of hacky, but since TypeScript doesn't support this kind of detail I'll merge :smile: You should prompt to author to instal use the `browser` field in `package.json` too - it'll result in smaller builds for their users.

PR #213, opened 2016-03-15
Add typings for knockout globals

SomaticIT:
Typings URL: [https://github.com/typed-contrib/knockout]
Source URL: [https://github.com/knockout/knockout]
Documentation URL: [http://knockoutjs.com]


SomaticIT:
I created a repository to maintain knockout typings in new repository but in order to have to both globals and module definitions I created two entries in registry.
Is it correct ?

blakeembrey:
@SomaticIT It depends, you may be able to use the global augmentation features in TypeScript or upcoming UMD features if that's what you need. But yes, it's fine to use both.

PR #216, opened 2016-03-16
Jest typings

asvetliakov:
Typings URL: [https://github.com/asvetliakov/typings-jest]
Source URL: [https://github.com/facebook/jest]

Existing DT jest typings are very bad


asvetliakov:
Looks like CI doesn't care about ambient definitions :smile:

blakeembrey:
@asvetliakov Are these ambient definitions? Want to put them under `global` or `env/` according to https://github.com/typings/registry#structure? Probably `env/`, although no ones using it yet.

asvetliakov:
Oh yes, i totally forgot :dizzy_face:

asvetliakov:
Should be global or env? Just checked global/ and looks like it already has some test specific stuff. What is `programming environment typings`? webpack-env typing from dt for example?

blakeembrey:
I'd use `env`, since isn't it a specific environment when running Jest and not just an import?

asvetliakov:
ok. you won :+1: 
moved to env

blakeembrey:
Haha! :+1: I was planning to fix it up myself, but never got around to it with all the other work I have ahead of me. Thank you :smile:

PR #218, opened 2016-03-19
Add typings for knockout.mapping module

SomaticIT:
Typings URL: [https://github.com/typed-contrib/knockout.mapping]
Source URL: [https://github.com/SteveSanderson/knockout.mapping]
Documentation URL : [http://knockoutjs.com/documentation/plugins-mapping.html]

### DIFF
--- a/npm/knockout.mapping.json
+++ b/npm/knockout.mapping.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "2.0.0": "github:typed-contrib/knockout.mapping#ea8c41fd6736738b8987f66187cb0b4851e0d15c"
+    }
+}


blakeembrey:
Wow, you're getting a lot of global/local typings in. I wonder if the TypeScript UMD feature will help you in 2.0.

SomaticIT:
For many cases it will.
But I thought about it and in this particular case, it will not. 
Because the module have a special behavior which will not be available in UMD feature.

Indeed, it is a normal module in AMD/CommonJS case but it does not exports any global in browser mode. It extends the Knockout module. I'm not sure that it will be doable even in TypeScript 2.0

PR #226, opened 2016-03-21
Add env-mocha

### LINKED ISSUE #81: mocha

unional:
Typings URL: [E.g. http://github.com/typed-typings/env-mocha]
Source URL: [E.g. https://github.com/mochajs/mocha]

Is it correct that I skip the `main` field when specifying `files`?
```js
{
  "name": "mocha",
  "ambient": true,
  "files": [
    "lib/global.d.ts",
    "lib/mocha.d.ts"
  ],
  "homepage": "https://github.com/mochajs/mocha"
}
```

Fixes #81


blakeembrey:
You _could_ still have a `main`, but it's optional. In your case, definitely unnecessary.

PR #231, opened 2016-03-22
support for markdown-it

rapropos:
Typings URL: [https://github.com/rapropos/typed-markdown-it]
Source URL: [https://github.com/markdown-it/markdown-it]

### DIFF
--- a/npm/markdown-it.json
+++ b/npm/markdown-it.json
@@ -1,6 +1,6 @@
 {
 	"versions": {
-		"4.0.0": "github:rapropos/typed-markdown-it#455e8241ab764385182fffa03dd06cd2080b2331"
+		"4.0.0": "github:rapropos/typed-markdown-it#71e0f4babb2878267f54503ed89ad05609df71f3"
 	}
 }
 
--- a/npm/markdown-it.json
+++ b/npm/markdown-it.json
@@ -1,0 +1,6 @@
+{
+	"versions": {
+		"4.0.0": "github:rapropos/typed-markdown-it#455e8241ab764385182fffa03dd06cd2080b2331"
+	}
+}
+


blakeembrey:
@rapropos Your type definition won't work according to the module definition. It uses the default CommonJS export, but you've used a named export (https://github.com/rapropos/typed-markdown-it/blob/master/index.d.ts#L100). Also, according to the source, it should be functions on the prototype (https://github.com/markdown-it/markdown-it/blob/master/lib/index.js#L577) while you've just exported them on the function. I think there's a bit of refactoring required to make this correct.

blakeembrey:
I think the main changes are:

```
declare function markdownIt (...): markdownIt.MarkdownIt;

declare namespace markdownIt {
  export interface MarkdownIt {
    ...
  }
}

export = markdownIt; // <- This export style is important here.
```

rapropos:
Thank you for taking the time to comment. I am struggling a bit to understand what is going on, so I will just describe what I tried and what happened:

* I tried doing exactly as you suggested, with only `MarkdownIt` inside the `markdownIt` namespace. This made all the other interfaces unimportable.

* So I put all the other interfaces (`Token`, `Rule`, &c) inside the `markdowtnIt` interface as well, which solved that, but now the constructor in my client code (which used to work) doesn't any more:

```
import {MarkdownIt} from 'markdown-it';

this.md = new MarkdownIt('default', {
                breaks: true,
                linkify: true,
            })
            .enable(['newline']);
```
I figure I have to change my import syntax, so I do this:
```
import * as markdown from 'markdown-it';
this.md = new markdown.MarkdownIt('default', {
                breaks: true,
                linkify: true,
            })
            .enable(['newline']);
```
...but no luck. 
```
Property 'MarkdownIt' does not exist on type '(preset?: string, options?: MarkdownItOptions) => MarkdownIt'.
```
I guess I can't construct `MarkdownIt` objects with `new()` any more? That seems sort of...suboptimal.

unional:
@rapropos , do a quick test: does this work?
```js
import MarkdownIt = require('markdown-it');
this.md = new MarkdownIt(...);
```

unional:
> Property 'MarkdownIt' does not exist on type '(preset?: string, options?: MarkdownItOptions) => MarkdownIt'.

This shows `import * as markdown from 'markdown-it';` gives you the function.
I then assume you are using `module: "commonjs"`, right?

blakeembrey:
@rapropos Did the code actually work before or did the compiler just not error? Because the compiler will believe what you say the types are, but the source code would not have worked before based on the examination I did. That's why I suggested changing the exports, so they properly match the source code.

rapropos:
@blakeembrey I guess I can't say for sure, but the webpack bundle looks this:
```
	var markdown = __webpack_require__(/*! markdown-it */ 579);
...
	var Markdowner = (function () {
	    function Markdowner() {
	        this.md = new markdown.MarkdownIt('default', {
	            breaks: true,
	            linkify: true,
	        })
	            .enable(['newline']);
...
```
...which looked OK to me. Now I'll switch back to doing it with the namespace and check @unional 's test.

Yes, I am using commonjs. Here are the compiler options in tsconfig.json:
```
 "compilerOptions": {
    "target": "ES5",
    "allowNonTsExtensions": false,
    "module": "commonjs",
    "sourceMap": true,
    "isolatedModules": false,
    "noEmitOnError": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "moduleResolution": "node",
    "noImplicitAny": true
  }
```

No, the test fails. I get three errors:
```
Cannot find name 'MarkdownIt'.
Only a void function can be called with the 'new' keyword.
'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.
```

If I change the type of `this.md` from `MarkdownIt` to `markdown.MarkdownIt`, the first error goes away but the other two remain. The second one is what inspired me to write the constructor function as returning void in the first place.

I apologize to both of you for not really knowing what I'm doing here.

blakeembrey:
@rapropos It's all good, I'll make a PR to your project and let's go from there :smile:

rapropos:
Updated commit hash to match the status after merging [@blakeembrey's assistance](https://github.com/rapropos/typed-markdown-it/issues/1).

PR #234, opened 2016-03-23
Adding ramda typings

enriched:
I don't think that these are absolutely complete.  I have been adding to them as I have found functionality that is missing and I grabbed them from https://github.com/donnut/typescript-ramda

Typings URL: http://github.com/enriched/typed-ramda
Source URL: https://github.com/ramda/ramda

### DIFF
--- a/npm/ramda.json
+++ b/npm/ramda.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "0.19.1": "github:enriched/typed-ramda#413395cd890d7f3236ced0a9718da9a3869840ec"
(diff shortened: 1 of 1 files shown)


blakeembrey:
Wow, that's a lot of typings. Cheers! :+1:

unional:
I think it is an ambient script, need to set `ambient: true` in `typings.json`?

blakeembrey:
Yeah, when TS 2.0 with UMD is released we'll also need to ping authors to support it :smile:

PR #235, opened 2016-03-23
Update README.md

johnnyreilly:
It's arguable whether you need to take this. But I thought it was worth being clear that `debug` in this context refers to an npm package. I hadn't heard of it myself and first read this thinking that debug was referring to the activity of debugging. So I was confused

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 ## Contributing
 
 1. Fork the project and run `npm install`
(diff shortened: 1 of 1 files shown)


blakeembrey:
Good point, that didn't occur to me :+1:

johnnyreilly:
It's the little things :smile:

PR #237, opened 2016-03-24
Add react-addons-update

nevir:
Typings URL: https://github.com/convoyinc/typed-react-addons-update
Source URL: https://github.com/facebook/react/blob/master/src/addons/update.js


blakeembrey:
LGTM :+1:

PR #238, opened 2016-03-26
Add registry entry for google-web-signin

flawless2011:
Typings definition for Google Web Sign-In library.  This is a migration from DefinitelyTyped so let me know if anything looks out of wack.

Typings URL: [https://github.com/flawless2011/typings-google-web-signin]
Source URL: [https://developers.google.com/identity/sign-in/web/reference#gapiauth2]

### DIFF
--- a/global/google-web-signin.json
+++ b/global/google-web-signin.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:flawless2011/typings-google-web-signin#5dabf489e0d3e64506a141ce8876439dd15560e5"
+  }
+}


unional:
You didn't expose it in "external format declaration". 
You can take a look at here: https://github.com/typings/typings/blob/master/docs/examples.md
Let me know if you need help.

blakeembrey:
@unional That's correct, it's meant to be global. Googles libraries are not external modules, but globals. It's also in the `global/` directory.

blakeembrey:
Awesome, thanks! :+1:

unional:
Ouch! Sorry~~

PR #240, opened 2016-03-26
Add typings for opn module

SomaticIT:
Typings URL: [http://github.com/typed-contrib/opn]
Source URL: [https://github.com/sindresorhus/opn]


blakeembrey:
Happy to merge, but you can re-use the promise interfaces from places like `es6-promise`. Although it's not an actual dependency that mirrors the NPM package, I'd promote you to do it since you're just re-using types.

PR #241, opened 2016-03-26
Add typings for jimp module

SomaticIT:
Typings URL: [https://github.com/typed-contrib/jimp]
Source URL: [https://github.com/oliver-moran/jimp]

### DIFF
--- a/npm/jimp.json
+++ b/npm/jimp.json
@@ -1,0 +1,5 @@
+{ 
+    "versions": { 
+        "0.2.20": "github:typed-contrib/jimp#54fdabe612ec0443d01f84b71efeeb9b101fbb9c" 
+    } 
+} 


blakeembrey:
Thanks! I'd remove this:

```ts
declare global {
    interface NodeRequireFunction {
        (id: "jimp"): typeof Jimp;
    }
}
```

It's tightly coupled to an implementation detail of the current `node.d.ts` file in DT, something we should avoid doing as you are breaking interop with other possible implementations. I don't actually see the need to do it either, since `import` semantics already handles this.

SomaticIT:
Fixed and merged.

I also removed the embed `Promise` typings in favor to `env~es2015-promise` which is not needed when targetting `ES6` in TypeScript.

blakeembrey:
@SomaticIT `jimp` does rely on a polyfill for promises anyway, so it's good to use that (https://github.com/oliver-moran/jimp/blob/master/package.json#L28).

PR #247, opened 2016-03-29
Add typed gulp-html-angular-validate

tdbrian:

Typings URL: https://github.com/tdbrian/typed-gulp-html-angular-validate
Source URL: https://github.com/ewu02/gulp-html-angular-validate

Thanks for contributing!


blakeembrey:
@tdbrian I think you want to unwrap https://github.com/tdbrian/typed-gulp-html-angular-validate/blob/master/index.d.ts from being inside `declare module`. This won't work when you install it otherwise.

PR #207, opened 2016-03-14
Adding typings for mongodb

CaselIT:
Typings URL: [https://github.com/Think7/typings-mongodb]
Source URL:[https://github.com/mongodb/node-mongodb-native]
Documentation URL:[http://mongodb.github.io/node-mongodb-native/2.1/api/]

PR #249, opened 2016-03-29
Mongodb updated commit hash

CaselIT:
The definition was updated. This commit updates the hast.
Original pull request of this type definition https://github.com/typings/registry/pull/207


blakeembrey:
Cheers! :+1:

PR #250, opened 2016-03-29
Adding typings for Riot Games' API

Protectator:
Hi ! I made type definitions to match Riot Games' API.

Typings URL: https://github.com/Protectator/league-typedef
Source URL: https://developer.riotgames.com/api/methods

As this is an API and not an actual js library, I tried to write types for multiple purposes. These definitiions are more aimed to be used in a server-side application that requests the API with the "Operations" interface I added in each module, which contains the available operations to the API as functions. But the type definitions can also be used in a client-side application, in which case using the other interfaces directly might be more useful.


blakeembrey:
@Protectator That's pretty interesting, I'd love to get this into the registry. Can we make a new directory, maybe using `lib/` or `etc/` to store these types of non-programming type definitions? It's a really cool use-case, I just don't want to mix it all up with everything else right now.

Protectator:
Sure, I just moved it.

blakeembrey:
Thanks :+1:

PR #253, opened 2016-03-29
added rxjs-symbol-typings

david-driscoll:

Typings URL: https://github.com/david-driscoll/rxjs-symbol-typings
Source URL: https://github.com/ReactiveX/rxjs/blob/master/spec/es5.d.ts

This is a small supplemental interface shim, to allow RxJS5 to target ES6 while still being compatible with TypeScript compilers that target `ES5`.  Only users that target ES5 will need to install this side by side with Rxjs5.

### DIFF
--- a/lib/rxjs-symbol-typings.json
+++ b/lib/rxjs-symbol-typings.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:david-driscoll/rxjs-symbol-typings#b06ca4cc670b9e0f2ace68a339dce38648e10e6c"
+  }
+}


blakeembrey:
Want to make a directory called `lib/` and use that? It'd be good to try and keep some structure of what the things are typed for. The name, I suppose, is pretty arbitrary?

david-driscoll:
It is fairly arbitrary, and can be changed if needed.  The big thing is it doesn't lead to any specific library.

Would having it under `/lib` cause any conflict, as it's supposed to be an ambient module?

david-driscoll:
file moved, and pr rebased from master.

blakeembrey:
Thanks :+1: Nope, `lib/` is meant to be ambient (https://github.com/typings/registry#structure). I think we can start storing the random ambient pieces under there so the README may need updating.

PR #257, opened 2016-03-31
adds typing for react-selectize

vintem:

Typings URL: https://github.com/furqanZafar/react-selectize
Source URL: https://github.com/furqanZafar/react-selectize

Thanks for contributing!

### DIFF
--- a/npm/react-selectize.json
+++ b/npm/react-selectize.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.9.0": "github:furqanZafar/react-selectize#d8bc4ee16a3a9dd0496b0c9f334c4b0204d7985b"
+  }
+}


unional:
Where is the typings? Seems like you are pointing to the source?

blakeembrey:
I see them https://github.com/furqanZafar/react-selectize/blob/develop/src/index.d.ts. @vintem, what you probably want to do instead is add `typings` to `package.json` - see http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html

vintem:
Hi @blakeembrey, what I really want to do is to have the typings available to the typings tool. I've posted a question about creating typings and I understood that I could do that in the original project.

[](https://github.com/typings/discussions/issues/13)

I guess that's not the case. Do I really have to create another project just for the typings? The project react-selectize has the typings.json on its root. That's not enough? 
Thanks for the help

blakeembrey:
@vintem No, you absolutely can. In this case, since you're distributing the the definition yourself, you can have users do `typings install npm:react-selectize`. Is that reasonable? If you want TypeScript users to use the definitions, it'd be better to allow them to do nothing. E.g. it just works, which is what happens when you use `package.json` instead.

The only thing you need to fix, in either case, is https://github.com/furqanZafar/react-selectize/blob/develop/src/index.d.ts#L94. You need to unwrap this module declaration and make it an external module declaration. E.g. Replace `declare module '...' { export = ... }` with just `export =`.

vintem:
Thanks for the help! I think I understand now, so the typings project is just for the cases where the original project does not host the typings file.

blakeembrey:
Yes :+1:

PR #274, opened 2016-04-01
Added definitions for pg v4.5.1.

mxl:
Typings URL: [https://github.com/goenning/typed-pg]
Source URL: [https://github.com/brianc/node-postgres]

PR #275, opened 2016-04-01
pg v4.5.1

goenning:
updating definition to pg 4.5.1


goenning:
Oops, this is a dup of #274

PR #277, opened 2016-04-01
Added is-thirteen@2.0.0

Fank:
Typings URL: https://github.com/Fankserver/typings-npm-is-thirteen
Source URL: https://github.com/jezen/is-thirteen


unional:
Although it doesn't matter in your case, but you can consider using `declare namespace` instead of `declare module`. :rose:

Fank:
sure let me quickly change it

PR #283, opened 2016-04-03
Added express-http-proxy@0.6.0

Fank:
Typings URL: https://github.com/Fankserver/typings-npm-express-http-proxy
Source URL: https://github.com/villadora/express-http-proxy

@unional I used generator-typing for the first time, can you validate i used it right?


unional:
Sure! Looks good. The source testing is still a work in progress and it is not trivial in this case.

One thing you can change is set `typings.json/version` to `0.6.0` to keep track locally which version you are typing for.

The `README.md` generated is from an older generator with wrong info. It should be `typings.json/version` instead of `package.json/version`.
 
Besides that, LGTM. :+1:

PR #288, opened 2016-04-04
Update bluebird.json

louy:
Adding version 3.3.4

### DIFF
--- a/npm/bluebird.json
+++ b/npm/bluebird.json
@@ -1,5 +1,6 @@
 {
   "versions": {
-    "3.2.1": "github:typed-typings/npm-bluebird#524d5cc66654fe7c8dcbccf90faeaa70588853a0"
+    "3.2.1": "github:typed-typings/npm-bluebird#524d5cc66654fe7c8dcbccf90faeaa70588853a0",
+    "3.3.4": "github:louy:typed-bluebird#9d1352843f7720fea2d91d01a9da0f97755ffb95"
   }
 }
--- a/npm/bluebird.json
+++ b/npm/bluebird.json
@@ -1,6 +1,5 @@
 {
   "versions": {
-    "3.2.1": "github:typed-typings/npm-bluebird#524d5cc66654fe7c8dcbccf90faeaa70588853a0",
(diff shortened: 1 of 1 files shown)


unional:
There is a typo. Also, do you want to update it in https://github.com/typed-typings/npm-bluebird?

louy:
Fixed. Thanks! Well I wouldn't mind, but I've been working on this fork for a long time now. We can probably merge the two. Would you make me a contributor to that repo?

unional:
@blakeembrey , can you add @louy to that repo? Lost ownership when transferred.

blakeembrey:
@louy I'll add you to `typed-typings`. Done.

blakeembrey:
@louy Done. What did you want to do with this PR now?

louy:
I'll update it once github starts working again! thanks.

louy:
@blakeembrey it's updated.

unional:
@Louy seems like you have other commit mixed with this

Also, should the previous version be kept?

louy:
I think squashing would solve this when you merge. You're right about previous version.  Sorry about that.

blakeembrey:
@unional What's the issue with the previous version? I don't see an issue with keeping it around?

unional:
Yes, my point is to keeping it around.

blakeembrey:
Ok, we should keep old versions :smile: The API is currently deprecating dead versions anyway, so if you do that it'll start emitting deprecated notices.

unional:
> so if you do that it'll start emitting deprecated notices.

Which bring up a point that if the user is using a specific version (e.g. `chai@3.4.0`), if `chai@3.5.0` is available would it print the deprecating warning?
I think it should be suppressed.

blakeembrey:
It's not marked as deprecated if you haven't deleted it from the registry, so it would be "suppressed". However, if you delete the version from the registry doesn't that constitute no more support for that version?

unional:
> constitute no more support for that version?

But why? If the typings is correct for that version, I don't see a reason to not support it if the user choose to use that particular version.

blakeembrey:
Because you deleted the version?

unional:
Yup. That's why I don't see much reason in removing entries in `typings/registry`, just like you normally won't `npm unpublish`. :smile:

PR #291, opened 2016-04-05
Add `normalizr@2.x` typings

unindented:
Typings URL: https://github.com/unindented/typed-normalizr
Source URL: https://github.com/gaearon/normalizr

### DIFF
--- a/npm/normalizr.json
+++ b/npm/normalizr.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.0.0": "github:unindented/typed-normalizr#58c96c9f2f765becaf531ded8990d7d3c72fb12c"
+  }
+}


blakeembrey:
@unindented Are you a maintainer of `normalizr`? Would you be interested in keeping the type definitions inside the repo itself instead? Doing that would provide immediate value for TypeScript users instead of making them use Typings.

unindented:
@blakeembrey You are right, that's the better option. Adding a `typings` field to the `package.json` that points to the `.d.ts` file would be enough, right?

blakeembrey:
Yes, that's all you need to do and TypeScript users will appreciate it :smile:

blakeembrey:
You're welcome to ping me too, if you need a definition review.

unindented:
@blakeembrey I've created a PR, if you have the time to review: https://github.com/gaearon/normalizr/pull/84

PR #292, opened 2016-04-05
Create electron.json

unional:

Typings URL: [E.g. http://github.com/typed-typings/env-electron]
Source URL: [E.g. https://github.com/electron/electron]

Thanks for contributing!


Fank:
@unional i think the test won't green up :weary:

unional:
Seem seems like some issue with github.

Fank:
LGTM i think updates will follow soon

PR #296, opened 2016-04-06
add global lodash

unional:

Typings URL: [E.g. http://github.com/typed-typings/npm-lodash]
Source URL: [E.g. https://github.com/lodash/lodash]

Thanks for contributing!

### DIFF
--- a/global/lodash.json
+++ b/global/lodash.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "4.0.0": "github:typed-typings/npm-lodash/global#e17b5e028baa7cc2400385c96c5a212b69c3756d"
+  }
+}


blakeembrey:
@unional Is there something stopping us from doing the `declare global` trick here?

unional:
Do you think the `declare global` trick will work if it is being referenced multiple times?

I recall it will lead to `duplicate identifier` error. But need to validate.

louy:
IMO, a global module should never be referenced multiple times. I don't like the idea of having a typing file for a global var that the user will have to expose manually anyway. 

moreover, module authors should never expose anything globally. think about having a module that uses lodash while another uses underscore.

unional:
These are for supporting user using the module through script tag

unional:
Merge since the augmentation trick is not working.

PR #297, opened 2016-04-07
Added filename.js typings

faceleg:

Typings URL: https://github.com/66pix/filename.js-typings
Source URL: https://github.com/66pix/filename.js

Thanks for your time!

### DIFF
--- a/npm/filename.js.json
+++ b/npm/filename.js.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "1.0.13": "github:66pix/filename.js-typings#4231c733e6863a3fdf39db3930c217552678e90f"
+        "1.0.15": "github:66pix/filename.js#733a5a700f715c4617e0b9d86d5c335fef1f9a07"
     }
 }
--- a/npm/filename.js.json
+++ b/npm/filename.js.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "1.0.13": "github:66pix/filename.js-typings#4231c733e6863a3fdf39db3930c217552678e90f"
+    }
+}


Fank:
Is it possible to place the typings in the source repository instead of a new one?

faceleg:
Yes of course, I wasn't sure if that was preferred (we could note this in the readme?)

faceleg:
Would I put the `node.d.ts` and `browser.d.ts` files at the root of the filename.js repo or in a typings folder?

Fank:
@blakeembrey or @unional can help

faceleg:
OK thanks. I'll wait for instructions before mangling anything else.

unional:
You and put is as `index.d.ts` and point to it through `package.json/typings`

For the content, you should not wrap them inside `declare module` just export them at top level

Seems like your nose and browser version are identical in api (briefly looked at it through phone). So you don't need two versions either. 

Hope it helps!

unional:
Using `export =` syntax as in your code

unional:
http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html

unional:
You can also configure this for your browser consumption:
https://github.com/blakeembrey/popsicle/blob/master/package.json#L12-L18

faceleg:
That's helpful, thanks. Just started with TS last night, appreciate the tips.

faceleg:
Done - I see github has a "squash commits" option when merging now, let me know if you want me to squash manually anyway.

unional:
As you have added it to the package itself, you don't need this PR. tsc will get the typings automatically.

faceleg:
Hooray I'm learning!

PR #301, opened 2016-04-08
Added jointjs typings

CaselIT:

Typings URL: [E.g. https://github.com/CaselIT/typings-jointjs]
Source URL: [E.g.  https://github.com/clientIO/joint]

Thanks for contributing!

### DIFF
--- a/global/jointjs.json
+++ b/global/jointjs.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.9.7": "github:CaselIT/typings-jointjs/global/typings.json#c6975d4a48cd34c44844cfa13f61319d143d3bc9"
+  }
+}
--- a/npm/jointjs.json
+++ b/npm/jointjs.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.9.7": "github:CaselIT/typings-jointjs#c6975d4a48cd34c44844cfa13f61319d143d3bc9"
+    "0.9.7": "github:CaselIT/typings-jointjs#b291e45a148ec9c0257b5bd7b38c213235de226b"
   }
 }
--- a/npm/jointjs.json
+++ b/npm/jointjs.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.9.7": "github:CaselIT/typings-jointjs#c6975d4a48cd34c44844cfa13f61319d143d3bc9"
+  }
+}


unional:
Seems like it is exposing:
> JointJS library exports three global variables: `joint`, `V` and `g`.

Would it be better to put `dia` `ui` `shapes` `util` and `layout` under `joint`?
Those will be place at the global namespace right now.

CaselIT:
No it exposes also the objects `dia` `ui` `shapes` `util` and `layout`

unional:
Oh, interesting. Never know it works that way. Thanks.

unional:
It's ES6 -> CommonJS interop

CaselIT:
I'm not sure I follow. Do you mean the missing exports in the namespace declaration?

unional:
Nope. I tested it and it is working.
`jointjs` is using `export =` syntax at `index.js`, and what you have is in `ES6 syntax`. So it is a ES6 -> CommonJS interop.

unional:
Not sure if that is a good idea to do so, as it might backfire as the CommonJS -> ES6 interop issue.
I think there are some ES6->CommonJS interop trick `babel` is doing.
But I guess this would be fine.

@blakeembrey , what do you think?

unional:
@CaselIT you probably aware of it, but for reference, what I'm talking about is related to:
https://github.com/Microsoft/TypeScript/issues/7398

CaselIT:
I was not aware of it, thanks for pointing it out. I think I did experience something similar while writing the type definition for this library. If load this library with system.js as global the external module as it is written is incorrect, while it works when loading it as commonjs. 
Still I think most people will use the ambient definitions and load the library with a script tag.
Do you have any advice to avoid the issue?

unional:
> Still I think most people will use the ambient definitions and load the library with a script tag

Loading from script tag just means the typings needs to be ambient/global/script. So the `/global/*` should take care of it.



> Do you have any advice to avoid the issue?

Currently no, unfortunately. That why I think the issue is massive.  The best we can do is to stay as close to the shape of the source as possible.

i.e. in this case keep it in the `export =` syntax.

CaselIT:
I tried to change the type definition, did you mean like they are now? They effectively are as before.

https://github.com/CaselIT/typings-jointjs/blob/test-export/index.d.ts

unional:
yup. :+1:

CaselIT:
I've updated the pull request.

unional:
FYI Having some difficulty to merge

CaselIT:
Can I help?

unional:
It's related to the repo setting I think. So need @blakeembrey's help on it.

CaselIT:
I can re-fork the repo and risubmit the pull request if you think that it would help.

unional:
Thanks. But it's not just your PR. Same for all other PRs. :smirk:

CaselIT:
:+1:

blakeembrey:
Sorry, it appeared to be an odd restriction that comes with the enablement of the squash option. It should now be fixed for collaborators.

PR #303, opened 2016-04-10
Add unambient immutable typings

alitaheri:
As it rose in https://github.com/typings/typings/issues/428 and the fact that immutable js is not being very actively maintained, I decided to add the unambient types to the registry, thanks to how the typings were originally written it was very easy! just remove 5 lines of codes and vola :sunglasses: I also renamed `module` to `namespace` as it's being deprecated. I tested it with my projects and it's working perfectly. please have a look.

Typings URL: https://github.com/alitaheri/immutable-unambient
Source URL: https://github.com/facebook/immutable-js

### DIFF
--- a/npm/immutable.json
+++ b/npm/immutable.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.7.6": "github:alitaheri/immutable-unambient#c7987e0b221760db96efccdf37222d24ed87759e"
+    "3.7.6": "github:alitaheri/immutable-unambient#3e5ef14e1677e2db90770b8263be889b71c4d444"
   }
 }
--- a/npm/immutable.json
+++ b/npm/immutable.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "3.7.6": "github:alitaheri/immutable-unambient#c7987e0b221760db96efccdf37222d24ed87759e"
+  }
+}


unional:
Thanks. It would be better to keep the shape in `export =` instead of ES6 named export style in this case because the source is UMD (i.e. `modules.export =` for npm).

Here are some relevant discussion:
https://github.com/typings/registry/pull/301#issuecomment-207853449

alitaheri:
I understand. Thanks for pointing it out :+1: :+1:

alitaheri:
@unional I minimized the modifications. Both to address the issue you mentioned and to make it easier to maintain. Right now, all I do, is turn:

```js
declare module "immutable" {
  export = Immutable
}
```

into:

```js
export = Immutable
```

unional:
Yup!

myitcv:
@alitaheri @unional given the ImmutableJS package, as distributed on `npm`, [includes a type definition](https://github.com/facebook/immutable-js/blob/master/dist/immutable.d.ts) (on which https://github.com/alitaheri/immutable-unambient is based) I don't think merging this PR with a definition from a repo that is unlinked to the Facebook repo is the right thing to have done.

The ImmutableJS type definition was discussed before:

* https://github.com/typings/typings/issues/49
* https://github.com/facebook/immutable-js/issues/705

unional:
@Alitaheri can explain better. Basically it is because the source is not active and the format can't be used by typings.

myitcv:
@unional I don't think it's accurate to say ImmutableJS isn't being maintained. It's a stable product, hence changes are infrequent. But it is maintained.

Given the ImmutableJS repo is the source of truth for the type definition, I'm just concerned:

1. we have now introduced another "source of truth"
2. there is no issue in ImmutableJS that is tracking a "proper" fix to this (unless I've missed it @alitaheri?)
3. @alitaheri now need's to ensure this "fork" is kept up to speed with (the admittedly slow-moving) ImmutableJS definition

Maybe the practical solution for now is to create an issue that addresses point 2 above

unional:
Sorry, not meant to say it is not being maintained. As you say, will just take time to get it in. :)

alitaheri:
@myitcv there were 2 reasons why I decided to go with this approach:

1. modifying typings provided by immutable-js is a breaking change. The `main` field in the package.json points to dist/immutable.js and there is a dist/immutable.d.ts which is ambient. messed up situation.
2. the stable version 3.7.6 cannot be modified to include the fixed typings. even if it was actively maintained, there has to be a _next_ version from which the fixed typings can be published from.

I had no choice, even if the typings could be fixed, It had to be fixed for the next version. now if some one is using, say version 2, there is no way they can have a fixed version of that typings residing in the oroginal repo, ulness for every major stable release there is a patch release solely for that purpose. granted that turning an ambient module into... a module? would be a breaking change requiring a major bump. It's twisted in too many ways that I thought to myself, come on just remove 2 lines of code, push to github and be done with it. :laughing:

myitcv:
@alitaheri 

> modifying typings provided by immutable-js is a breaking change.

Yes, I agree it is a breaking change. 

> the stable version 3.7.6 cannot be modified to include the fixed typings

I agree, because `v3.7.6` has been released

> even if it was actively maintained, there has to be a next version from which the fixed typings can be published from

Have you discussed on https://github.com/facebook/immutable-js whether this is an option? 

> I had no choice

I don't think it's accurate to position this as the only solution. Your `next` solution is one, there are others.

> It's twisted in too many ways that I thought to myself, come on just remove 2 lines of code, push to github and be done with it

But this solution introduces other problems:

1. we have now introduced another "source of truth"
2. this second source of truth (your "fork") has to manually be kept up to speed with (the admittedly slow-moving) ImmutableJS definition
3. we haven't addressed the issue of multiple versions of the type definition for different versions of TypeScript, discussed in https://github.com/typings/typings/issues/49

@alitaheri this isn't a personal criticism, more a comment directed towards @unional @blakeembrey et al on how we should be approaching this sort of situation.

alitaheri:
@myitcv Those are valid points. I should have discussed it with the core team as well. But, I would suggest you to have a look at https://github.com/typings/typings/issues/428 as well. This was the original reason why I even came across this concern. If ambient type definitions that are originated from npm packages were installed correctly (that is, not install at all since they are ambient) when installed as dependency of another dependency, we wouldn't even need to fix `immutable` type definitions. What I did here was a quick fix, so others wouldn't have to face this issue, it took me some hours to get to the bottom of it. I also tried fixing another issue with index files but the code needs research and I'm not familiar with it. So as soon as that one is fixed, and ambient types from npm won't cause trouble, I'll deprecate this.

blakeembrey:
@myitcv Preferably, we should be updating the package itself. However, I did look at the issue backlog and they don't seem to want to make any movement on fixes. Just see https://github.com/facebook/immutable-js/issues/734, it's over 4 months old, exactly this issue and zero contributor comments. In this situation, I suppose we can just open another issue for visibility.

blakeembrey:
Yeah, the PRs and issues there are a rabbit hole. Someone did the correct fix: https://github.com/facebook/immutable-js/pull/808/files. However, not a single one of these issues have any comments from collabs, I can not expect any of it to be maintained.

leebyron:
Hey, this came to my attention recently that these were lost in the noise. Will help get a release out.

blakeembrey:
@leebyron Awesome, cheers! Although that's the correct implementation, I'm sure it'd be easier on your build system to go the reverse direction - wrap it in the declaration. But I'm sure you'll fix that up 😄

leebyron:
Yeah, immutable-js's build system is a pre-babel mess that uses both gulp and grunt. Not my favorite thing. It will get cleaned up over time :)

alitaheri:
@leebyron Just saw the release, awesome work :+1: :+1:

PR #304, opened 2016-04-10
Add typings for env-node

### LINKED ISSUE #251: Typings for node
It'd be great to create the `env-node` based on Typings that used `typings.json` and dependencies.

louy:
Typings URL: http://github.com/typed-typings/env-node
Source URL: https://nodejs.org/

Closes #251

### DIFF
--- a/env/node.json
+++ b/env/node.json
@@ -1,0 +1,9 @@
+{
+  "versions": {
+    "0.8.0": "github:typed-typings/env-node/0.8#8feb3e2f7099efe714d44ef95562d4fba7a671c5",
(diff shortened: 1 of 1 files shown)


unional:
LGTM. :+1: 
One minor thing, in your `package.json`, you can add `"private": true` to avoid accidentally publishing it to npm.

louy:
I never publish packages on my machine to avoid such cases and use CI instead, but yeah good shout. I'll add that.

blakeembrey:
Wow 👍 Should it before `4.0` for consistency? I may help contribute some time, I know I've been meaning to fix the DefinitelyTyped definitions forever.

PR #305, opened 2016-04-11
added typings for agenda

cdbajorin:

Typings URL: [E.g. https://github.com/cdbajorin/typed-agenda]
Source URL: [E.g. https://github.com/rschmukler/agenda]

Thanks for contributing!


blakeembrey:
Awesome, LGTM. @unional The `source-test` copying omits having `devDependencies` - should that be part of the copy somehow?

unional:
> @unional The source-test copying omits having devDependencies - should that be part of the copy somehow?

Those detail is still being worked on. Likely yes, but some are not needed (e.g. linting).

PR #313, opened 2016-04-13
adding chai-xml

rhysawilliams2010:
Typings URL: http://github.com/rhysawilliams2010/typings-chai-xml
Source URL: https://github.com/krampstudio/chai-xml

### DIFF
--- a/npm/chai-xml.json
+++ b/npm/chai-xml.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.3.1": "github:rhysawilliams2010/typed-chai-xml#142be915fa7e0996f4cf48f28cce11901106efe4"
+  }
+}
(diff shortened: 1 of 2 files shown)


blakeembrey:
@rhysawilliams2010 I can't merge this because of the augmentation of the namespace hack instead of the real name. Unfortunately, I know that's currently the only way to make it work. It's been discussed in https://github.com/typings/registry/pull/270#issuecomment-204583615 and the fix is currently blocked by a compiler issue in TypeScript. I'll let you know when it's fixed and we can merge.

rhysawilliams2010:
I have changed the branch on my fork for this PR. There is a new PR here which i have already closed: https://github.com/typings/registry/pull/344 . I can reopen later when I can make it less hacky.

PR #316, opened 2016-04-15
add Hen typings

bang88:

Typings URL: [E.g. http://github.com/henjs/typed-hen]
Source URL: [E.g. https://github.com/henjs/hen]

Thanks for contributing!


blakeembrey:
@bang88 If you're the author of `hen`, you can move the type definitions inline. See https://github.com/typings/registry/pull/257#issuecomment-204115865.

PR #319, opened 2016-04-16
Update blue-tape.json

unional:
Temp revert to not releasing `0.2.0` to avoid bug introduced

### DIFF
--- a/npm/blue-tape.json
+++ b/npm/blue-tape.json
@@ -1,6 +1,5 @@
 {
   "versions": {
-    "0.1.0": "github:typed-typings/npm-blue-tape#3de9a90b45538efc80d9c87f4a51f99abb8e810b",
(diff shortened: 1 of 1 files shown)


unional:
@blakeembrey would downgrade works? Will it mess up indexing?

blakeembrey:
@unional It doesn't mess up indexing, feel free to.

unional:
Ok, at least I hope it doesn't affect other people. :sob:

blakeembrey:
@unional It's perfectly fine. I'll make a change to fix up the API server with deprecated typings now, I think I have the correct behaviour in mind and will release a new version with changes today. The change: `/entries/:source/:name/versions/latest` will only show non-deprecated versions, while `/entries/:source/:name/versions/:version/latest` will. Basically, let's you install specific unsupported versions when being explicit but not when you just want to install any version.

PR #324, opened 2016-04-16
Add npm/gulp-count

MrHen:
Typings URL: https://github.com/MrHen/typings-gulp-count
Source URL: https://github.com/palantir/gulp-count


unional:
option is missing `title`
Also in https://github.com/MrHen/typings-gulp-count/blob/master/index.d.ts#L4,
I think it can be written as:
`    logger?(message: string): any;`
to tell `tsc` it is an optional function instead of an optional variable of type function.

🌷

MrHen:
Updated with https://github.com/MrHen/typings-gulp-count/pull/2

PR #332, opened 2016-04-18
Added rwlock typings

CaselIT:

Typings URL: [E.g. https://github.com/CaselIT/typings-rwlock]
Source URL: [E.g. https://github.com/71104/rwlock]

Thanks for contributing!

### DIFF
--- a/npm/mongodb.json
+++ b/npm/mongodb.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.1.0": "github:Think7/typings-mongodb#3a50a13eeb9fbb63f8759200dcb2bf576ff33285"
+    "2.1.0": "github:Think7/typings-mongodb#e72e4c62896ceee80c8786f2e10b28da6c8c2481"
   }
 }
--- a/npm/rwlock.json
+++ b/npm/rwlock.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "5.0.0": "github:CaselIT/typings-rwlock#f3519a868b4a3a989e8376316b43b72b26d40e44"
+    "5.0.0": "github:CaselIT/typings-rwlock#40643e82365784176a7547412c17dd8863ed4a05"
   }
 }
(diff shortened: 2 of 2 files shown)


unional:
https://github.com/71104/rwlock/blob/master/package.json#L18
Seems like the main is `lock`, may be better to name your typings as `lock.d.ts`

unional:
Pure styling: since it is namespaced in `ReatWriteLock` already, you probably don't need to name the interfaces as `RW*` https://github.com/CaselIT/typings-rwlock/blob/master/rwlock.d.ts#L13

CaselIT:
Even if the import is with `rwlock`?
Regarding the `RW` I'll remove them.

unional:
> Even if the import is with rwlock

The general convention would be following what's in the `package.json/main` for "require-ability" in case there are multiple files. Doesn't really matter in this case. :smile:

@blakeembrey, I asked about this a while ago:

This case is clear:
```
// in source:
// package.json/main = "a.js" or "a"
// folder:
a.js
lib/b.js

// in typings:
// typings.json/main = "a.d.ts"
```

Would like to confirm on this:
```
// in source:
// package.json/main = "lib/a.js" or "lib/a"
// folder:
lib/a.js
lib/b.js

// in typings:
// typings.json/main = "a.d.ts" or "lib/a.d.ts"?
```

CaselIT:
I'll also rename the file then :+1: 
Should I also put it under `lib`?

unional:
I think you don't have to. You can change that in the future if needed. Will not do any harm since it is the main.

CaselIT:
Committed the changes :+1:

blakeembrey:
FWIW, you won't be able to access `ReadWriteGeneric` outside of the module. Hope that's ok.

@unional Let's stay true to the source. If anything, that's the source of truth and there needs to be no extra questions.

CaselIT:
`ReadWriteGeneric` is only used to avoid code repetition so there is no need to access it outside the module. That's why I defined it outside the namespace.

unional:
> @unional Let's stay true to the source.

So meaning keep it as lib/a.d.ts. :smile:

PR #335, opened 2016-04-19
Adding twit

Volox:
Typings URL: https://github.com/Volox/typed-twit
Source URL: https://github.com/ttezel/twit


blakeembrey:
@Volox Can you replace `NodeJS.ReadableStream` with the module usage? E.g. `import { ReadableStream } from 'stream'` and use that? It'll avoid it breaking unexpectedly if the magic global ever disappears (or is not implemented).

Volox:
@blakeembrey Done.

blakeembrey:
Awesome work!

PR #338, opened 2016-04-19
Update PR template

blakeembrey:
Keeping it really simple. Maybe we can add a checklist later of what the typings being added should look like? E.g. 

* [ ] Has simple README with link to source?
* [ ] Follows source repository structure?
* [ ] Is an external module (typings for library) or ambient (typings for global or env)?


Fank:
We should remove the E.g. in the url because so many people only replace the url but leave the E.g.

PR #339, opened 2016-04-20
jsonld.js typings

tpluscode:
Typings URL: https://github.com/tpluscode/jsonld-typings
Source URL: https://github.com/digitalbazaar/jsonld.js

### DIFF
--- a/npm/jsonld.json
+++ b/npm/jsonld.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.4.8": "github:tpluscode/jsonld-typings#e7f830a3224ea158a7ebfdfef2afa01c2f8fd8dc"
+  }
+}
--- a/npm/jsonld.json
+++ b/npm/jsonld.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.4.8": "github:tpluscode/jsonld-typings#e7f830a3224ea158a7ebfdfef2afa01c2f8fd8dc"
+    "0.4.0": "github:tpluscode/jsonld-typings#6859a022544a3091988974f7a0e9c32c9773d603"
   }
 }


blakeembrey:
@tpluscode I'm happy to merge this, but I'd consider refactoring the definition a little. Right now, you create a variable and only export that. It loses everything else. I'd convert to be more ES6 like so you can properly export all types. For example:

```ts
export function compact(doc:any, context:any, callback:JsonLdCompactCallback);
export function compact(doc:any, context:any, options:CompactionOptions, callback:JsonLdCompactCallback);

export function expand(compacted:any, callback:JsonLdCallback);
export function expand(compacted:any, options:ExpansionOptions, callback:JsonLdCallback);
```

blakeembrey:
Also, don't use `module` in an external module format - it's deprecated. (Ref: https://github.com/tpluscode/jsonld-typings/blob/master/jsonld.d.ts#L3).

unional:
> Also, don't use module in an external module format

Use `declare namespace` instead.

tpluscode:
Thanks for your suggestion. Please help me out here. I'd like to make it right :). 

I've updated my typings so that they work fine with how one can use the API. I've actually dropped the module/namespace completely. The only thing I don't like is that Webstorm suggests interfaces which weren't exported explicitly like `NormalizeOptions` etc.

AFICT there are a number of ways to consume jsonld.js. I do it by importing `promises`

``` js
import {promises} from 'jsonld';
promises.compact().then();
```

It is also possible to import and construct `JsonLdProcessor` (exposes same API as `promises`)

``` js
import {JsonLdProcessor} from 'jsonld';
new JsonLdProcessor().flatten().then();
```

And lastly there are the functions with callbacks in global scope.

``` js
import * as jld from 'jsonld';
jld.flatten(doc, ctx, () => {});
```

Interestingly any import also populated a global `jsonld` object. Even if it is `import 'jsonld'`. Is there anything to do about that?

blakeembrey:
@tpluscode The latest update looks good (from a brief look at it). Did you want to update the PR to include it?

> The only thing I don't like is that Webstorm suggests interfaces which weren't exported explicitly

This is normal, because usually you need access to the non-exported types in the exported types so for that to work they need to be exported. Try it with a `.ts` file and you'll see TypeScript complain about this.

> Interestingly any import also populated a global jsonld object.

What do you mean here? I think maybe it wasn't wrapped by Typings at the time you were playing with it?

tpluscode:
Sure, I'll update the PR unless there are some more changes to be made.

By the way the readme says:

> Use semver to specify the minimum version the typings are valid for in the registry

How is that version used when installing typings?

> What do you mean here? I think maybe it wasn't wrapped by Typings at the time you were playing with it?

Can't reproduce now so I guess it was some glitch in Webstorm.

blakeembrey:
> How is that version used when installing typings?

It's only informative for others during installation.

tpluscode:
Okay, merge away.

And regarding version, what happens when there are more than one?

blakeembrey:
> And regarding version, what happens when there are more than one?

As in, there's two in the registry? You can always install either by version, but it'll always install the latest by default.

PR #343, opened 2016-04-21
add sns-validator

rhysawilliams2010:

Typings URL: https://github.com/rhysawilliams2010/typings-sns-validator
Source URL: https://github.com/aws/aws-js-sns-message-validator

Thanks for contributing!


unional:
Seems like you can improve the `message: Object` a bit to be an interface. The values are fixed.

unional:
Also, in `typings.json`, the `version` should be the version of the source your are typings against, which is 0.1.0.
And it would be great if you include its `homepage`. That's what will show up when user do `typings search sns-validator`

PR #346, opened 2016-04-21
Adding cheerio

Volox:
Typings URL: https://github.com/Volox/typed-cheerio
Source URL: https://github.com/cheeriojs/cheerio


blakeembrey:
@Volox You might want to check out how to do branching, that way you can submit two PRs at once without causing issues (notice how this has your previous `twit` PR included and can not be automatically merged as a result).

PR #347, opened 2016-04-21
Add npm:facebook-chat-api

Demurgos:
Typings for `npm:facebook-chat-api`

Typings URL: https://github.com/insa-frif/typed-facebook-chat-api
Source URL: https://github.com/Schmavery/facebook-chat-api

### DIFF
--- a/npm/facebook-chat-api.json
+++ b/npm/facebook-chat-api.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "1.1.0": "github:insa-frif/typed-facebook-chat-api#db2780ef99705af4d568230c16893f3ea52d372c"
+    }
+}


blakeembrey:
Awesome, LGTM 👍 The only oddity, though it works fine, is https://github.com/insa-frif/typed-facebook-chat-api/blob/master/index.d.ts#L692-L708 - usually I just define a function with the same name at the top. In this case, they're functionally equivalent anyway 😄

Demurgos:
Okay, I already received some feed-back from the library author so I have some details to fix.

I used this pattern because it helped me before when a module exposed a function and this function also had some attributes. But you are right, it feels a bit strange so I'll define it as a simple function.

blakeembrey:
@Demurgos When it has attributes (properties on the function) you can define a namespace with the same name and the definitions merge. Feel free to submit a new PR when you've made changes 😄

PR #354, opened 2016-04-23
Added sweetalert 2 typings.

StefanKoenen:

Typings URL: https://github.com/StefanKoenen/typed-sweetalert2
Source URL: https://github.com/limonte/sweetalert2

Thanks for contributing!

### DIFF
--- a/npm/sweetalert2.json
+++ b/npm/sweetalert2.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.3.0": "github:StefanKoenen/typed-sweetalert2#3515ba54463eda307c77cddfa29eb3366aaf4147"
+  }
+}
--- a/npm/sweetalert2.json
+++ b/npm/sweetalert2.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.3.0": "github:StefanKoenen/typed-sweetalert2#3515ba54463eda307c77cddfa29eb3366aaf4147"
+    "1.3.0": "github:StefanKoenen/typed-sweetalert2#d6876d41c373fff49791b81125cb366ff1965e2d"
   }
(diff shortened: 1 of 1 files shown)


blakeembrey:
@StefanKoenen The only issue is that typings under `npm` should be in external module format, which means it needs to be tweaked a bit and use `declare global` instead to introduce global definitions on `window`.

StefanKoenen:
@blakeembrey Thanks for the feedback, i didn't know this.
I tryed to fix it, but i don't know how. After i changed it (the way i though it should be) i get the following error:
Attempted to compile "sweetalert2" as an ambient module, but it looks like an external module. Can you show me what i'm doing wrong?

(My code is at my github)

blakeembrey:
Don't use the `--ambient` flag, just install it regularly (Typings wraps it up for you).

blakeembrey:
The changes look correct, if you update the PR I'd be happy to merge 👍

StefanKoenen:
@blakeembrey Thanks, I've updated the PR.

I though the ambient flag was for libraries which are registered globally? I'm little bit confused now when to use the flag :)

blakeembrey:
Pretty much, yes. All the flag does is tell Typings that it shouldn't wrap up the definition to make it work. I'll get some better error handling in the ASAP.

StefanKoenen:
@blakeembrey How am i suppose to use this in my angular2 app now? 
it doesnt recognise swal as function...

blakeembrey:
@StefanKoenen You'll need to provide a gist or some example code, I've only been guessing how it's used and it sounds like my guesses are wrong. Most likely, you were actually relying on global interface merging hacks like those that exist in DefinitelyTyped.

PR #355, opened 2016-04-25
Added typings-golden-layout

stephenjelfs:


Fank:
source url and typings url please

stephenjelfs:
@Fank Here are the URLs--please let me know if I also need to commit these somewhere?

The project is golden-layout:
https://www.npmjs.com/package/golden-layout
https://github.com/deepstreamIO/golden-layout

And the typings are here:
https://github.com/Genedata/typings-golden-layout

blakeembrey:
Awesome, cheers! Have you considered using classes for the declarations instead of manual interfaces?

stephenjelfs:
@blakeembrey I replaced the GoldenLayout interface with a class:

[https://github.com/Genedata/typed-golden-layout/commit/972bae8d97678e580986bda305cf73029a67aa7f]

But, I was not sure if you also mean the other interfaces, inside the namespace, because they are not exported?

blakeembrey:
Yes, that's all I was currently referring to.

PR #357, opened 2016-04-26
Update bluebird.json

louy:
Added new version and deprecated old one due to a bug in the typings file.

Methods `.inspect` and `.settle` have been deprecated since version 2. Replaced with `.reflect`.


blakeembrey:
@louy Don't want to keep the old hash around?

louy:
nope. it had a bug.

PR #358, opened 2016-04-26
Add npm/js-sha256

Demurgos:
Typings URL: https://github.com/Demurgos/typed-js-sha256
Source URL: https://github.com/emn178/js-sha256

I added a warning in the `README.md` because I had a doubt: this module both uses the Node's "function as a module" export and exports a `namespace` being strict-commonjs/ES6 standard-compliant.
When both are available, should I still expose the non-standard way ? With Typescript supporting ES6 modules from the beginning, it's does not complicate the import on the consumer side.

This way of exporting also has a few side-effects: the object is circular:
````typescript
import {sha256} from "js-sha256";
console.log(sha256.sha256 === sha256); // true
````
But this is not documented.
Also, the implementation uses an optional parameter but the documentation never mentions it: it's in fact an internal "flag". Should I define it as optional or do not expose this private signature ?

### DIFF
--- a/npm/js-sha256.json
+++ b/npm/js-sha256.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.3.0": "github:demurgos/typed-js-sha256#ce991a5c69d819c4ee7a6de223b1233d2bf16fb7"
+  }
+}


blakeembrey:
I think you've documented it perfectly 👍

blakeembrey:
I'm actually not sure how you could represent the circular structure without duplicate type definitions right now, and since it's undocumented I think you're safe 😄

Demurgos:
I did not try it but I think that I could have done it this way: 

````typescript
interface JsSha256 {
    (message: string | Array...): string; // computes the sha256 of the message
    sha256: JsSha256; // circular reference
    sha224(message: string | Array...): string; // computes the sha224 of the message
}
declare let sha256: JsSha256;
export = sha256;
````
should be equivalent to:
````typescript
sha256.sha256 = sha256;
sha256.sha224 = sha224;
module.exports = sha256;
````

But as I told you, I was not sure if `typings` should encourage best practice/standard when it's compatible or should the definition be "hacky" (It's how I feel about `export = ...`, just compare the above code with the one I submitted: it's less obvious what happens)  and strictly follow the real interface ? 
I already expressed in an other issue that I agree that it should follow the original module (ie. for the project structure) for compatibility reasons, but here the export/imports that are already ES6 compatible.

blakeembrey:
I think the existing code is fine, as I said. I don't see a need for it to be complicated that way, unless it's how people use it (and here, it's not)

Demurgos:
Ok, thank you for you answer.

PR #359, opened 2016-04-26
Update npm/facebook-chat-api

Demurgos:
Typings URL: https://github.com/insa-frif/typed-facebook-chat-api
Source URL: https://github.com/Schmavery/facebook-chat-api

Changes:

 - `facebookChatApi.getUserInfo`

 - Normalize type for ID (when an argument is an ID, it is `number | string`, an ID in a result is always `string`)

- Errors are not native errors but `ErrorObject`s

- Use the method proposed in #347 to export the main function

- Lint and minor re-factoring


blakeembrey:
Cheers! 👍

PR #372, opened 2016-04-28
Added registry entry for Id3-parser

dvdcxn:
**Typings URL:** https://github.com/dvdcxn/typed-id3-parser

### Questions (for new typings):

**Q: Does the README explain the purpose and a link to the typed source?**

Y

**Q: Do the typings follow the source structure**

I tried to replicate existing typings but not 100% sure i've done it appropriately.
 
**Q: Are they correctly external or ambient modules?**

They should be although again I'm not 100% sure of the difference between the two. Just learning TS atm, sorry about that.

### Typing description

Added registry entry for Type definition for creeperyang's [id3-parser](https://github.com/creeperyang/id3-parser).

### DIFF
--- a/npm/id3-parser.json
+++ b/npm/id3-parser.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.1": "github:dvdcxn/typed-id3-parser#6f3b7442206a96eabd0f875cf20994783b2936f0"
+  }
+}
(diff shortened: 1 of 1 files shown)


blakeembrey:
@dvdcxn Unfortunately, there's a small change you'll need to make to the typings. Typings under `npm/` should be in external module format (which means there's an `export` or `import` at the top-level) and not the ambient module format. To fix that, you pretty much just need to remove the `declare module "id3-parser"` wrapper.

dvdcxn:
That's not unfortunate at all :) 

That's it changed now, thanks for the suggestion and explanation!

blakeembrey:
Thanks! 👍

PR #375, opened 2016-04-28
Add meteor typings to the registry

barbatus:
**Typings URL:** https://github.com/barbatus/meteor-typings

Add type definitions for Meteor 1.3. 
There was a PR to the DefinitelyTyped regarding this (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/9113) but I thought it's time to move definitions here.
I believe it will be maintained in barbatus/meteor-typings (or under another global repo in the future) from now on.

### DIFF
--- a/env/meteor.json
+++ b/env/meteor.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.3.0": "github:barbatus/meteor-typings/meteor/1.3#83c6933d0ef07abfd4a2a85b5ec41bc3949f6d28"
+  }
+}


blakeembrey:
Awesome. I'm not that familiar with Meteor, will you need to use the `main` and `browser` typings at the same time ever?

barbatus:
@blakeembrey If I understand it right, and `main.d.ts` is supposed to be definitions for the NodeJS env, then yes - Meteor has API available only on the server side.

PR #376, opened 2016-04-28
Add npm/npmlog

Demurgos:
**Typings URL:** https://github.com/Demurgos/typed-npmlog

I don't have any particular remarks for these typings. It's not perfect yet (some methods are dynamically created) but I'm starting to get used to write definitions.

### DIFF
--- a/npm/npmlog.json
+++ b/npm/npmlog.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.0.3": "github:Demurgos/typed-npmlog#8f62e95b2a949806242aa684221feba1b79b909a"
+  }
+}
--- a/npm/npmlog.json
+++ b/npm/npmlog.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.0.3": "github:Demurgos/typed-npmlog#8f62e95b2a949806242aa684221feba1b79b909a"
+    "2.0.3": "github:Demurgos/typed-npmlog#406c8b6bb3757bae77e2b2e6e56d19008b45c228"
   }
 }


Demurgos:
I just spotted some errors with these definitions. I will quickly fix them and update this PR (in like 10 minutes).

Demurgos:
Ok, just when I thought that I was getting used to typings, I got an error related to the dynamically generated methods.

Here is the related doc:

> `log.addLevel(level, n, style, disp)`
> 
> - `level` {String} Level indicator
> - `n` {Number} The numeric level
> - `style` {Object} Object with fg, bg, inverse, etc.
> - `disp` {String} Optional replacement for level in the output.
> 
> Sets up a new level with a shorthand function and so forth.
> 
> Note that if the number is Infinity, then setting the level to that will cause all log messages to be suppressed. If the number is `-Infinity`, then the only way to show it is to enable all log messages.
> 
> _[...]_
> 
> `log[level](prefix, message, ...)`
>
> Like log.log(level, prefix, message, ...). In this way, each level is given a shorthand, so you can do `log.info(prefix, message)`.

I tried to use the following declaration:
````typescript
interface NpmLog {
  // level
  // stream
  // log (...)
  // addLevel (...)
  // ...
  [level: string]: (prefix: string, ...message: any[]) => void;
}
````
But it caused errors because the last part was incompatible with the other properties. I did not hear about a way to define a "fallback" type for properties, but maybe there is a way around this problem ?

Demurgos:
Even if the dynamic methods are not typed yet, this PR should already cover most of the needs. It defines the signatures for the dynamic methods added by default. (`info`, `silly`, `warn`, etc.).

blakeembrey:
@Demurgos Thanks, happy to merge. On the fallback, the signature has to be compatible so they only way I can think of is to make it loose enough that the rest work - or maybe you could return an intersection of both types (haven't tested this).

Demurgos:
Ok, thank you. If I encounter this situation again (and I think that I will...), I'll post an issue on the Typescript repo to get some help.

blakeembrey:
Sure, no worries. Happy to help when it comes to writing definitions, but I'm sure more people are watching there anyway 😄

PR #381, opened 2016-04-30
typings for annyang

theluk:
**Typings URL:** https://github.com/theluk/typed-annyang


blakeembrey:
@theluk Just an FYI, since you're only using a single namespace and exporting it there's actually no need for the namespace at all. For example, you currently have:

```ts
declare namespace x {
  export function test (): boolean;
}

export = x;
```

That's the same as doing:

```ts
export function test (): boolean;
```

At the top level instead.

PR #382, opened 2016-04-30
Update node.json

louy:
Added some props & methods to Error and ErrorConstructor


blakeembrey:
@louy Did you want access to the registry?

louy:
That would be very handy!

PR #388, opened 2016-05-02
Would you verify is everything correct, please. My apology if something is not in proper order - I'm new in TypeScript and JavaScript.

pr0peller:
Typings URL: [https://github.com/pr0peller/jquery-pagings-typings]

### DIFF
--- a/global/jquery-paging.json
+++ b/global/jquery-paging.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.2.0": "github:pr0peller/jquery-pagings-typings#b981e571b85cc481f9ed73d87fdb4f84af6911bb"
+    "1.2.0": "github:pr0peller/jquery-pagings-typings#5fbc0618c14ffb1d9344c627f08f54fb47c0dba7"
   }
 }
--- a/global/jquery-paging.json
+++ b/global/jquery-paging.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.2.0": "github:pr0peller/jquery-pagings-typings#6259a366f2831d6c7e544ee3edc6aad598539bfc"
+    "1.2.0": "github:pr0peller/jquery-pagings-typings#19c535be5ca040b899280a01c346ec681030efc4"
   }
 }
--- a/global/jquery-paging.json
+++ b/global/jquery-paging.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.2.0": "github:pr0peller/jquery-pagings-typings#6259a366f2831d6c7e544ee3edc6aad598539bfc"
+  }
+}
--- a/global/jquery-paging.json
+++ b/global/jquery-paging.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.2.0": "github:pr0peller/jquery-pagings-typings#19c535be5ca040b899280a01c346ec681030efc4"
+    "1.2.0": "github:pr0peller/jquery-pagings-typings#b981e571b85cc481f9ed73d87fdb4f84af6911bb"
   }
 }
--- a/npm/jquery-paging.json
+++ b/npm/jquery-paging.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.8.0": "github:pr0peller/jquery-pagings-typings#a274170dac91c19e7a20a770400f14f0805fa8a5"
+    "1.2.0": "github:pr0peller/jquery-pagings-typings#a274170dac91c19e7a20a770400f14f0805fa8a5"
   }
 }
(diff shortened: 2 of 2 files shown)


blakeembrey:
Unfortunately things in `npm/` should be in external module format - if you want to write a global module you can put it in `global/`. Aside from that, you can also kill the module declaration to correct it (https://github.com/pr0peller/jquery-pagings-typings/blob/dev/index.d.ts#L76-L78).

pr0peller:
Hi,

I have removed the module and variable declarations. I red about modules but steel I'm not sure that i understand the concept of global and external module. Would you verify please.

pr0peller:
If I understood correct, module declaration must be removed and all interfaces have to be exported.

blakeembrey:
Hey! Sorry I didn't get back to you earlier, do you happen to be using the jQuery typings from DefinitelyTyped?

pr0peller:
Hi, I use jQuery from typings not from DefinitelyTyped.

blakeembrey:
Oh, sorry, I believe they still come from DefinitelyTyped then 😄 

```
jquery                        dt     http://jquery.com/                                                    1        2016-04-17T21:32:36.000Z
```

So, with that in mind, have you tested this definition? I believe it'll cause some problems because it relies on interface merging with the global jQuery interface. Sorry to mess you around, it's just that's the difference between `--ambient` and without. Unless we get a working external module version of jQuery, you might want to use a global version of this module and put it in `global/` instead.

pr0peller:
I actually think about this, and I thing that it is better to put in global/. I think that is the best. Wold you please verify just the index.d.ts file?

blakeembrey:
@pr0peller How do you import it? Do you need to create a module so `import 'jquery-paging'` works?

pr0peller:
I was using that as in the first (original) commit with the "declare module" and just `/// <reference path="..."` in the beginning of the file.

blakeembrey:
Right, but `declare module 'x'` wouldn't work with normal definitions - global definitions aren't wrapped up though, so if you're relying on the import working you'll need it again. Sorry for the confusion.

pr0peller:
What I did is I looked from ambient definition for "jstree" (installed from typings) and i did this the same way. It was working for me and decided to share it with other people. I'm new to typescript and JavaScript (I'm coming from C#) and those module things are very confusing for me. For a week now, as you can see, I'm struggling to figure out what is the way to export module in description so other people can access it and use it. If I follow you, what you were saying i have to export the interfaces which I want to be accessible. Em I right sir?

blakeembrey:
Yes.

blakeembrey:
But only if it's a non-ambient definition. The difference is that `--ambient` installs verbatim, while without it wraps it up for TypeScript to understand. Since you want to write an ambient (or global) definition, you just need to declare the interfaces. The reason I asked if you import it using `import ''` or if it exists globally already is because of how you'd use it. It's a bit of a mess, but if you need `import 'foo'` to work you need a `declare module 'foo' {}` to exist.

blakeembrey:
Feel free to share a direct use-case of how you're using the library in JavaScript and we can figure out how to make it work in TypeScript too (especially if nothing I'm saying makes sense, it's not your fault).

pr0peller:
I think the problem is that this is jQuery plugin and it extends jQuery by adding `.paging` function to it.
I have declare all the interfaces. This should work for as ambient definition. I'll take a look again, when my knowledge and understanding abort those things is better and stronger. Thanks for the explanation. Deeply appreciated you patient. Thank you.  :)

blakeembrey:
No worries. You can always do test installations locally to check if it works, might be easier to test that way too 😄 (if you're not already).

The current implementation _will_ work (I believe), but you won't be able to do `import 'x'` (because there's no module declaration). As long as you aren't doing that, then it's fine. There **a lot** of different ways to load JavaScript because there's been no standard for a long time - E.g. `<script />` tags (which I guess you're doing now), ES6 imports (`import 'x'`) which TypeScript transpiles to other implementations (because no one supports it fully yet), CommonJS (synchronous `require('x')`), AMD (asynchronous `define(['x'], function (x) {})`), and other custom implementations.

pr0peller:
Thank you. I'm trying to use RequireJS because it make sense for me. and it works good with TypeScript. Steel a lot to learn though. Thank you one more time :)

blakeembrey:
Personally I'd recommend just sticking with CommonJS - there's a much bigger ecosystem and less issues to worry about. With AMD, you can only use AMD or UMD modules - which is a very small ecosystem and blocks out most of the NPM ecosystem. That's just my advice though 😄 I know it can be a pain to get your head around all the ways you can do things in JavaScript.

PR #400, opened 2016-05-07
Create mongoose.json

louy:
**Typings URL:** https://github.com/louy/typed-mongoose

### DIFF
--- a/npm/mongoose.json
+++ b/npm/mongoose.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "4.3.5": "github:louy/typed-mongoose#9d7b28ad7689659e908439c2fbf00d2510777c12"
+    "4.3.5": "github:louy/typed-mongoose#8d6b5aeb91b33ee901300254e490495f5c0d27a5"
   }
 }
--- a/npm/mongoose.json
+++ b/npm/mongoose.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "4.3.5": "github:louy/typed-mongoose#0e81218c59476fc0d95efb5ddd560a4b8d4d807d"
+    "4.3.5": "github:louy/typed-mongoose#9d7b28ad7689659e908439c2fbf00d2510777c12"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
@louy Why not use a promise dependency? Does `mongoose` require a global promise to exist?

louy:
[Mongoose uses](http://mongoosejs.com/docs/promises.html) [`mpromise`](https://www.npmjs.com/package/mpromise). But you can override it by setting `mongoose.Promise` to whatever value you want. 

I'm not sure how to model that in typescript in any way other than using the global promise. Plus `mpromise` isn't typed yet.

blakeembrey:
If it's just a standard `Promise`, just add a dependency on `es6-promise` and use that. Then people can just work, it's what I usually do (since they're all ES6 compliant).

louy:
It's compliant but not standard. They have all those weird methods like `chain/end/fulfil/onFulfil` that no one probably uses.
But what about es6 targets? I know ambient dependencies don't get installed by default but I'm afraid this might change in the future. Will it ever change?

blakeembrey:
Doesn't matter in ES6 targets. That's a global vs a local. No, ambient dependency behaviour will never change.

I'd type it with the `es6-promise` for now. Maybe you can replace it later, but it's better than relying on a global version.

louy:
Makes sense. If anyone wants to use `mpromise` they're free to add typings for it. I've updated the repo.

blakeembrey:
@louy Sorry, I meant use the non-global version that's typed in our registry through NPM. It doesn't introduce globals and let's you use it by doing `import { Promise } from 'es6-promise'`.

blakeembrey:
Here's an example: https://github.com/typed-typings/npm-inquirer/blob/master/lib/inquirer.d.ts#L1. I think we should only rely on globals if the underlying library also does the same - that way, users get accurate compiler errors compared to their runtime.

stevehipwell:
Do we have an estimation as to when this will be added?

blakeembrey:
Once @louy gets back to me.

PR #412, opened 2016-05-13
Adding typings for pify

cyounkins:
**Typings URL:** https://github.com/cyounkins/typed-pify

### DIFF
--- a/npm/pify.json
+++ b/npm/pify.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.3.0": "github:cyounkins/typed-pify#d2e231b233884fe46ca5e65f670c0da8cbba1866"
+  }
+}


blakeembrey:
FWIW, it is possible to achieve more complex implementations using generics: https://github.com/typed-typings/npm-thenify/blob/master/index.d.ts.

Thanks! 👍

blakeembrey:
Do you want to at least make the return a function that returns a promise, instead of `any`?

cyounkins:
I would love to make the types better, but I messed with it for a long while and couldn't get it to work.

pify(function) = promise but pify(module) = module, but the types don't match because the methods of said module are all mutated. I looked at Bluebird.promisifyAll but it wasn't any better.

In the typings I have commented out definitions for zero and one-arg functions, but I couldn't get them to work in conjunction with module definitions. Perhaps you could try?

blakeembrey:
Promifying a single function is easy to provide type signatures for (see link above). A whole module is unfortunately impossible, because there's no way in TypeScript to iterate over keys of an object to define a new type. I'm happy to merge as is, in that case.

PR #416, opened 2016-05-15
Add type definitions for i2c-bus

101100:
**Typings URL:** https://github.com/101100/typed-i2c-bus

### DIFF
--- a/npm/i2c-bus.json
+++ b/npm/i2c-bus.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:101100/typed-i2c-bus#3af7588bedbac004de1bd65b445381bf6268e7f4"
+  }
+}


101100:
The typing uses `Buffer`, but I didn't want to tie to any particular version of node.js.  How should I handle that?

blakeembrey:
Just use `Buffer`, it's fine. If you want, list node as an ambient dependency and it'll warn for you.

Aside from that, make sure the definition is in external module format. E.g. remove `declare 'x'` (https://github.com/101100/typed-i2c-bus/blob/master/i2c-bus.d.ts#L5).

101100:
I've removed the `declare` wrapper.

How would I mark node as an ambient dependency?

blakeembrey:
@101100 You can do `typings install env!node --save --ambient` to install the ambient node typings. That really just creates an entry in `typings.json` for you and the key is what really matters in the warning.

101100:
OK, I understand now, I think.  I've added a note to the `README.md` that the user will need the node ambient typings installed.

Ready for another look.

PR #420, opened 2016-05-16
Create fuse.js.json

MikaAK:
**Typings URL:** https://github.com/MikaAK/fuse-typings/

- [x] Does the README explain the purpose and a link to the typed source?
- [x] Do the typings follow the source structure?
- [x] Are they correctly external or ambient modules?

Add typings for `fuse.js` since the typings on definitely typed are for `fuse` and do not work correctly. Should be also noted typings on DT are out of date and the [PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/8598) for an update is now 2 months old and doesn't take advantage of generics. 
## NOTE

I am getting errors from running `npm run lint+build+test` with the typings-generator i used. I am using these typings currently fine so I'm not sure what to do about these errors.

``` bash
./out/main.d.ts (9,13): Module augmentation cannot introduce new names in the top level scope. (2665)
../out/main.d.ts (26,13): Module augmentation cannot introduce new names in the top level scope. (2665)
../out/main.d.ts (31,3): Exports and export assignments are not permitted in module augmentations. (2666)
    at getOutput (/Users/mika/Documents/side/typings-fuse/node_modules/ts-node/src/ts-node.ts:210:13)
    at Object.loader (/Users/mika/Documents/side/typings-fuse/node_modules/ts-node/src/ts-node.ts:225:23)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at /Users/mika/Documents/side/typings-fuse/node_modules/tape/bin/tape:31:13
    at Array.forEach (native)
    at /Users/mika/Documents/side/typings-fuse/node_modules/tape/bin/tape:30:15
    at f (/Users/mika/Documents/side/typings-fuse/node_modules/once/once.js:17:25)
```

### DIFF
--- a/global/fuse.js.json
+++ b/global/fuse.js.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.2.0": "github:MikaAK/fuse-typings#2d6cae662908210123da474499e22a7e436bb9f7"
+  }
+}


unional:
it's because you are not writing it as a module. Remove the `declare module ...` wrapper

blakeembrey:
@unional This module is meant to be global, according to the directory it's in (FWIW, Typings is now correctly erroring again on global vs non-ambient options). 

@MikaAK Can you add `global: true` to `typings.json` instead? That'll bundle it as a global definition.

unional:
Oh I see. Will it be a bit more complicated then? Since the test still based on typings 0.x, global won't work with `typings bundle`. I'll see if I can do a PR to fix it

unional:
Interesting, `fuse.js` use `vows` as it test framework. @blakeembrey, do you know how to run `ts-node` with it? I can open an issue on `ts-node`

unional:
Or open an issue on `vows` to add a `--require` feature. :)

unional:
PR filed https://github.com/MikaAK/fuse-typings/pull/1

MikaAK:
Merged the PR, I'm not fully familiar with ts import syntax and don't really understand the difference between es6 `import {x} from 'some-module'`  and `import x = require('some-module')`. Via assistance I found `import x = require('some-module')` would work with my typings vs the es6 syntax not working at all. I played around with it for a few hours but at the end I gave into the `import x = require(...)` syntax.

unional:
THat's about loader interop. Basically avoid it. If `some-module` is `commonjs` module, use `import x = require(...)`. If it is es6 module, use `import x from 'some-module'` or `import {x} from 'some-module'` depends on what does it export.

MikaAK:
@unional any idea how i can avoid it? It's not a global typing and `import x from 'some-module'` throws errors!

unional:
`import x = require('some-module')`

MikaAK:
Sorry I haven't had the chance to revisit this till now.
Using `import Fuse from 'fuse.js'` provides me with `No default Export`. 
Using `import * as Fuse from 'fuse.js'` returns `Module ''fuse.js'' resolves to a non-module entity and cannot be imported using this construct.` 
Using `import {Fuse} from 'fuse.js'` returns ` Module ''fuse.js'' resolves to a non-module entity and cannot be imported using this construct.`

unional:
`import Fuse = require('fuse.js')` is the older typescript syntax for commonjs module. 
They did some interop between commonjs and es6 so that you can use `import * as Fuse from 'fuse.js'`.
However there are issue with it and currently IMO we should avoid it.

See:
https://github.com/unional/typescript/blob/master/style-guide/default/modules.md#import-keyword

unional:
As for how to avoid `Module ''fuse.js'' resolves to a non-module entity and cannot be imported using this construct`, need to spend some time looking into the typings.

MikaAK:
I no longer use this lib, I'm not sure what's needed!

PR #428, opened 2016-05-19
Add Freewall definitions

Maddosaurus:
**Typings URL:** https://github.com/Maddosaurus/typed-freewall


unional:
Since you are defining `global` module, you should set `typings.json/global` to true.

Maddosaurus:
I updated `typings.json` in  the typings-freewall repo and adapted the ref in `freewall.json`.

PR #433, opened 2016-05-19
Add typings for npm/base64.json

twaldecker:
**Typings URL:** https://github.com/twaldecker/typings-base64


unional:
It does not follow source structure. The source main is `base64.js`, so your typings should be `base64.d.ts` instead of `index.d.ts`

unional:
Also, it would be better to add `version` and `homepage` in `typings.json`

blakeembrey:
A better way to write this may also be:

```ts
export function encode (str: string): string;
export function decode (str: string): string;
export const version: string;
```

PR #434, opened 2016-05-19
Add typings for npm/blob-util

twaldecker:
**Typings URL:** https://github.com/twaldecker/typings-blob-util

- [x] Does the README explain the purpose and a link to the typed source?
- [x] Do the typings follow the source structure?
- [x] Are they correctly external or global modules?


blakeembrey:
@twaldecker Did you test the definition? I think this will be an issue: https://github.com/twaldecker/typings-blob-util/blob/master/index.d.ts#L2. You have an interface by not type.

PR #439, opened 2016-05-24
Add define file for leancloud-jssdk.

wujun4code:
**Typings URL:** https://github.com/leancloud/typed-leancloud-jssdk

* [ ] Are they correctly external or global modules?

For more info : [LeanCloud , to be the best BaaS service provider](https://leancloud.cn)

### DIFF
--- a/npm/leancloud-jssdk.json
+++ b/npm/leancloud-jssdk.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0-rc8": "github:leancloud/typed-leancloud-jssdk#3f54847861139fbe98fe3931bf0a2282dc7b7c31"
+  }
+}
--- a/npm/leancloud-jssdk.json
+++ b/npm/leancloud-jssdk.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0-rc8": "github:leancloud/typed-leancloud-jssdk#839134b3e590ec34b16ad3fff288ddd23928ea66"
+    "1.0.0-rc8": "github:leancloud/typed-leancloud-jssdk#026dc965bcdd9f4c10b7a65181c27c69626927fb"
   }
 }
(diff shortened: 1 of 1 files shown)


wujun4code:
here is the tyings.json of the project  -> [typed-leancloud-jssdk](https://github.com/leancloud/typed-leancloud-jssdk/blob/master/typings.json)

```js
{
  "name": "typed-leancloud-jssdk",
  "main": "index.d.ts",
  "homepage": "https://github.com/leancloud/typed-leancloud-jssdk",
  "version": "1.0.0-rc8",
  "globalDependencies": {
    "node": "registry:dt/node#4.0.0+20160319033040",
    "blue-tape": "registry:npm/blue-tape#0.1.0+20160322235613"
  }
}
```

who can tell me what can i do to fix this problem?

blakeembrey:
If you're adding a module to `npm`, it needs to be in external module definition format. You've written it as a global definition. To fix this, you really just need to unwrap it from `declare module '...` as seen https://github.com/leancloud/typed-leancloud-jssdk/blob/master/index.d.ts#L6. This is the third checkpoint and the source of your error.

wujun4code:
Thanks for comments.
But,where can I get a sample for update from 0.8 to 1.0.x ?
I searched a lot, but when I try to unwrap the module, the compiler popup an error.

blakeembrey:
This has nothing to do with 0.8 or 1.0. You've written the definition wrong for Typings. You need to do what I said above.

blakeembrey:
The wrapping up of the file into a global definition is Typings' job.

wujun4code:
Thanks a lot. 
I will fix this issue.

blakeembrey:
Thanks. I'm happy to merge. FWIW, I think you may have ties to the original author? You can actually author the definitions as part of the module itself and skip using Typings altogether. That would probably be a better path in the future, and if that happens feel free to move the entry from the registry.

wujun4code:
3ks.
I am the original author.I am not familiar with typings define file.But I tested all the d.ts in an ionic project.so I want to provide it to other developers.
BTW,typings is an awesome tool.I wil make more define files for TypeScript.

PR #440, opened 2016-05-25
Update node.json

unional:
For https://github.com/typed-typings/env-node/issues/10

Does this look right?


blakeembrey:
Looks correct. I think the test script needs to be updated too 😄

unional:
Would this fix fine with you?

blakeembrey:
I guess that works for now 😄

PR #441, opened 2016-05-25
Add a base ES2015 env typings to unify modules depending on ES2015 features (Promise)

SomaticIT:
**Typings URL:** https://github.com/typed-contrib/es2015
**Specification:** http://www.ecma-international.org/ecma-262/6.0/

I made this typings to unify modules dependencies. This avoid to:
* Avoid the use of different "polyfill" definitions in each modules (Promise, Map, Set, etc.).
* Allows users to use ES2015 features on their ES5 targetted projects or lib.
* Allows users to take only the features they need.

Do you think it's a good idea ?

### DIFF
--- a/env/es2015-array.json
+++ b/env/es2015-array.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:typed-contrib/es2015/array/typings.json#0.1.3"
+  }
+}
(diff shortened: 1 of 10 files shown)


SomaticIT:
❓

blakeembrey:
@SomaticIT I'd love to make sure at least one other person reviews PRs for new type definitions, or give 48 hours. I've just been busy lately, but I did want to say that this concept is perfect to me 👍

PR #442, opened 2016-05-25
Add nightmare.json

JumpeiArashi:
**Typings URL:** https://github.com/JumpeiArashi/typed-nightmare
**Source URL:** https://github.com/segmentio/nightmare/

nightmare.d.ts doesn't exist in typings.

* create nightmare.d.ts for version 2.x (with electron version)
* updated typings to 1.0.x on generated by `typings-generator`
* pass travis ci test

See https://github.com/JumpeiArashi/typed-nightmare/blob/master/index.d.ts for detail. Thanks!

### DIFF
--- a/npm/nightmare.json
+++ b/npm/nightmare.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.4.1": "github:JumpeiArashi/typed-nightmare#5d3c06ebfb9a10c02f836ed91e339866adffd591"
+  }
+}
--- a/npm/nightmare.json
+++ b/npm/nightmare.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.4.1": "github:JumpeiArashi/typed-nightmare#0b23e780781b1c2836913c661d76103967f5bcfb"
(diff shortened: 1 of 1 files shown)


blakeembrey:
@JumpeiArashi Looks great. You may want to put all those extra interfaces into the `Nightmare` namespaces so people can still access them outside your typings.

Side note: What exactly is the `electron` namespace?

JumpeiArashi:
Are _all those extra interfaces_ `NightmareOptions`, `Paths` and `Switches` ? Would it be better to expose those Interface?

> What exactly is the electron namespace?

Nightmare is wrapped electron. So some Object (mainly Interface as nightmare arguments) are electron's. So I defined electron namespace to separate into Nightmare original interface for readability. Would it be better to re-define as Nightmare interface ,too?

blakeembrey:
Yeah, it just let's people re-use those interfaces. Sometimes really helpful - especially when you want to use that types to work with on your own objects. Thanks 😄

JumpeiArashi:
Thank you for merging:metal: Surely, I understood benefits of exporting interfaces. Thanks.

PR #455, opened 2016-05-26
Definition for base64-js 1.1.2.

mxl:
**Typings URL:** https://github.com/mxl/typed-base64

* [v ] Does the README explain the purpose and a link to the typed source?
* [v ] Are they correctly external or global modules?

Source library has structure `lib/b64.js`, but I put it to index.d.ts. Is it ok?


blakeembrey:
Sure, no worries. FWIW, the definition could be even simpler - there's no need for the namespace. Just do:

```ts
export function toByteArray(b64: string): Uint8Array;
export function fromByteArray(uint8: Uint8Array): string;
```

PR #460, opened 2016-05-28
Update Meteor typings 

barbatus:
https://github.com/meteor-typings/meteor

Move typings  to a new repo under meteor-typings organization.

Add typings for Meteor 1.2, which is ambient.

Typings for Meteor 1.3 now just defines new namespaces introduced in 1.3 which depends on globals from Meteor 1.2.

### DIFF
--- a/env/meteor.json
+++ b/env/meteor.json
@@ -1,6 +1,5 @@
 {
   "versions": {
-    "1.3.0": "github:meteor-typings/meteor/1.3#64f0d62cb163049fd2c04c71ac1a8402b4ff6c9e",
     "1.2.0": "github:meteor-typings/meteor/1.2#64f0d62cb163049fd2c04c71ac1a8402b4ff6c9e"
   }
 }
--- a/env/meteor.json
+++ b/env/meteor.json
@@ -1,5 +1,6 @@
 {
   "versions": {
-    "1.3.0": "github:barbatus/meteor-typings/meteor/1.3#83c6933d0ef07abfd4a2a85b5ec41bc3949f6d28"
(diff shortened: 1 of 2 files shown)


blakeembrey:
@barbatus The error seems to be that everything written as 1.3 typings appear to be external modules instead of global. Was that intentional?

barbatus:
@blakeembrey 1.2v and 1.3 share the same modules. But, while in 1.2v they are true ambient (i.e., `Mongo`), 1.3v has same the module as `meteor/mongo`. Here is description https://github.com/meteor-typings/meteor. The real point though is that I didn't find simpler way to share modules other then make them external in 1.3v.

barbatus:
@blakeembrey should I move 1.3v to the lib registry?

barbatus:
Not sure why it's failing, locally `npm test` shows no error.

blakeembrey:
@barbatus `lib`, `global` and `env` are all meant to hold global typings (see README). There's `common` which would work for external typings like this one, but you're welcome to rename it as no one has used it yet.

barbatus:
@blakeembrey I just removed `ambient` from `typings.json`, and tests have passed locally. But it's still as you see failing here, not sure where is discrepancy

blakeembrey:
@barbatus It should be failing. Why it's not locally, I can't really answer. Try running tests on all files instead of just the last commit? I'm going to guess the last commit you have is actually valid, but the entire branch is not.

barbatus:
@blakeembrey fixed

blakeembrey:
Thanks! 👍

PR #462, opened 2016-05-30
Create when.json

louy:
**Typings URL:** https://github.com/typed-typings/npm-when

### DIFF
--- a/npm/when.json
+++ b/npm/when.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.4.0": "github:typed-typings/npm-when/2#12702e1a1ef8ebcca68e5b50a9304304444dad76"
+  }
+}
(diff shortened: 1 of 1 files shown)


blakeembrey:
LGTM 👍 I think you could also use the `es6-promise` typings, if you wanted to make it consistent with other ES6 Promise typings implementations.

louy:
I don't like the idea of using `es6-promise` for everything. It's not accurate at all and might allow someone to make mistakes. e.g. `when` doesn't have `Promise#race`

blakeembrey:
@louy `Promise#race` is part of the spec for promises, so to be spec compatible I believe it would have it. I saw this: https://github.com/cujojs/when/blob/master/docs/api.md#whenrace.

blakeembrey:
@louy If there's any inaccuracies in the typings though, definitely mention it.

PR #463, opened 2016-05-31
Add type definitions for riot

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-riot

### DIFF
--- a/npm/riot.json
+++ b/npm/riot.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.3.0": "github:effervescentia/typed-riot#0b70513564d3982fdd5cbd211ff30696b9ac8d97"
+  }
+}


unional:
Seems like you are missing a `typings.json` under the `global` folder.
Also remember to add `global: true` in that file.

effervescentia:
Thanks :) all better now

unional:
Good. btw, is there a reason not offering a module version (under npm/riot.json instead of global/riot.json)?

effervescentia:
I hadn't intended on publishing it to npm as it seemed like unnecessary overhead and I presumed `typings` only relied on the `typings.json` file. I don't mind versioning & publishing to npm if that's the preferred practice

unional:
I see. The `npm` in registry doesn't mean publishing to `npm`. It means "the package is from npm". i.e. if you use `riot` from `'npm install riot` and then `import riot = require('riot');`, then it is better to create and use the `npm/riot.json` version as it does not pollute the global namespace.


I've created a PR on your repo. You can merge and add the `npm/riot.json` 🌷

effervescentia:
When I run the tests now it complains that `riot` ought to be a global module

unional:
Ok. Let me check.

blakeembrey:
@effervescentia typings under `npm/` should be written as external modules, while typings under `global/` can be written as global modules. The folders are meant to be consistent, which is why you're now getting the error. 

If you look at `dist/` (the path you've tried to add to `npm/` - https://github.com/effervescentia/typed-riot/blob/master/dist/index.d.ts) you'll see it's published as a "global module". That means it's wrapped in the `declare module '...'`.

unional:
Yeah, just point to the root, not `dist`

effervescentia:
@unional okay, but at the same time shouldn't the result of `typings bundle -o <filename>` result in a module rather than a global module?

unional:
The answer to that is a bit more complicated. It is best described here:
https://github.com/typings/typings/blob/master/docs/external-modules.md

Essentially, once you bundled, it becomes immutable and includes all dependencies.
Imagine you use a npm module which depends on `lodash` and embed the whole `lodash` in its distribution.

effervescentia:
@unional I'm no longer able to reference the types from the `Riot` namespace, is this because it's a `module` now instead of `global`? I assume it being wrapped with `declare module` means (as you put before) that the types are no longer globally accessible. As an example, I've included the definitions using `typings` in a separate project I'm working on and can neither reference the namespace `Riot` nor the inner `Observable`. I have `import riot = require('riot);` at the top of the file, and it references the `typings` index correctly using the triple slash path reference: `/// <reference path="../typings/index.d.ts" />`. Any ideas?

unional:
Added test PR for you.

unional:
If you are writing in modules, you probably should not be accessing namespace.

effervescentia:
@unional Here's the code I'm working with. I want to be able to specify that some variable (`emitter` in this case) has a certain type so that I can autocomplete to the various `Riot.Observable` functions. I've attached a screenshot, could you recommend the correct implementation that gives me access to certain interfaces that I want to reference from my code?

screen shot 2016-05-31 at 16 21 56

unional:
I take another look into your typing. I think there are other problems. Will try to fix it later on.

unional:
When you say `Riot.Observer`, you are accessing the global namespace. The types should be accessible from the imported module, i.e., directly in `riot`.

effervescentia:
This makes much more sense 😄 Thank you @unional (and @blakeembrey), you've been most helpful

PR #467, opened 2016-05-31
Added type definition of headless-gl

LimeStreem:
**Typings URL:https://github.com/jThreeJS/typed-headless-gl** 

### DIFF
--- a/npm/headless-gl.json
+++ b/npm/headless-gl.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "4.0.1": "github:jThreeJS/typed-headless-gl#71821673b53cee52e3006a6483ddfd1b72cfa40e"
+  }
+}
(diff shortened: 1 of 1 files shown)


LimeStreem:
What is the differences between global module and external module?
This definition works well in my local environment....

How should I fix this?

blakeembrey:
See https://github.com/typings/typings/blob/master/docs/faq.md#types-of-typings.

blakeembrey:
@LimeStreem You have two big bugs. One, it should be in external module format (remove the `declare module '...'` from wrapping everything. The other bug is that `export default` will not work here - it should be `export =`.

LimeStreem:
Thank you a lot!
I would not be able to fix this without your comment.

blakeembrey:
All good. A technique some of us use to when doing `export =`-style typings is to create a namespace and put the interfaces inside. This allows us to access the interfaces outside of the definition correctly. You can see an example here: https://github.com/typings/typings/blob/master/docs/examples.md#function--utility

PR #474, opened 2016-06-01
add npm/react-select

zaytsevvla:
https://github.com/zaytsevvla/typed-react-select

* [ ] Are they correctly external or global modules?

### DIFF
--- a/npm/react-select.json
+++ b/npm/react-select.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:zaytsevvla/typed-react-select#e63fcc5dd40ce35c8c6c148d04d24a8532ddbc75"
+    "1.0.0": "github:zaytsevvla/typed-react-select#3f75176e4f59612e6bad27246af8f577d048dc2d"
   }
 }
--- a/npm/react-select.json
+++ b/npm/react-select.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:zaytsevvla/typed-react-select#61b223854951a45f732910f1279a23276a6ae744"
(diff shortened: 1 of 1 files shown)


zaytsevvla:
` Attempted to compile "react" as an external module, but it looks like a global module`
please give me an example of using global module

unional:
Your dependency is wrong. The dt~react should be a global.

I saw a PR on react coming in. Do you want to use that instead?

blakeembrey:
@zaytsevvla https://github.com/typings/typings/blob/master/docs/external-modules.md and https://github.com/typings/typings/blob/master/docs/faq.md#types-of-typings.

zaytsevvla:
`I saw a PR on react coming in. Do you want to use that instead?` - nope
but please, let me know when it would be merged, i'll fix this typing.

blakeembrey:
@zaytsevvla Can you change https://github.com/zaytsevvla/typed-react-select/blob/master/lib/Select.d.ts#L1 to `import from 'react'` instead? `dt~react` isn't actually a module, it's `react` "from" `dt` (DefinitelyTyped).

blakeembrey:
Also, is https://github.com/zaytsevvla/typed-react-select/blob/master/typings.json#L10 working for you somehow? That shouldn't actually work currently, I'd recommend making it a `globalDependency` or maybe even a `peerDependency`.

zaytsevvla:
@blakeembrey used `peerDependency` as in package and `import from 'react'`. look like it works
But i can't understand why `dependencies` + `import from 'react'` didn't work.

blakeembrey:
@zaytsevvla Because the `react` definitions you are using are written in "global" module format, not as external modules. Typings does not use the global sub-dependencies, they are used for client hints.

blakeembrey:
By the way, it's `peerDependencies`, but I'm happy to merge as is and you can correct it later when needed 😄

PR #489, opened 2016-06-02
Added v4.0 to react-router-redux

CarsonF:
Latest definition changes to [react-router-redux#caf792c](https://github.com/andrew-w-ross/typings-react-router-redux/commit/caf792cb79efbf54ce328366809dc52821192a73) are for v4.0.
Updating here so it can be used :)


blakeembrey:
Thanks! 👍

PR #491, opened 2016-06-03
the https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/ch…

wujun4code:
**Typings URL:** https://www.npmjs.com/package/chart.js

…artjs/chart.d.ts is deprecated.

### DIFF
--- a/npm/chart.js.json
+++ b/npm/chart.js.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.1.4": "github:wujun4code/typed-chart.js#72dad16171c405f1d91065afedee74b46475c384"
+  }
+}
--- a/npm/chart.js.json
+++ b/npm/chart.js.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.1.4": "github:wujun4code/typed-chart.js#d4e2557805ebedabe77456ef4286a6e79c9f2cd9"
+    "2.1.4": "github:wujun4code/typed-chart.js#6d86e2c801cfd55c452842546d1b3bef91525462"
   }
 }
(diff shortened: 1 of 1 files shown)


wujun4code:
I will fix it.

blakeembrey:
@wujun4code I'm sure you realised the issue, but it's the fact that https://github.com/wujun4code/typed-chart.js/blob/master/index.d.ts has no exports - which makes it a "global" definition. You'll need to refactor it a bit to mirror `chart.js` as an external module.

wujun4code:
@blakeembrey thanks.
I was busy at the weekend.I will recode it to an  external module.

wujun4code:
@blakeembrey done
👌

blakeembrey:
@wujun4code Have you tried this against the original source code? From the docs, it looks like you may need to format the `.d.ts` in `export =` style.

wujun4code:
@blakeembrey Thanks again.
But I do not know how to wrap the Chartjs with `export = ` style.
Should I wrapped it as a `namespace` ? but when wrapped , the code maybe look like this：

```ts
namespacle Chartjs {
  class Chart {
      ....
  }
}
```

is it ok? Thanks.

blakeembrey:
So it'd end up looking like this:

```ts
declare class Chart {
}

declare namespace Chart { ... }

export = Chart;
```

There's a little example in the Typings docs here: https://github.com/typings/typings/blob/master/docs/examples.md#exporting-class--static-methods--utility

wujun4code:
thanks so much.
I will fix it.

wujun4code:
@blakeembrey I recoded it.
Thanks.

Issue #164, opened 2016-03-02
Noop module typings


blakeembrey:
Expose a TypeScript "noop" module for quickly getting around type errors. See https://github.com/typings/typings/issues/110#issuecomment-191064423 for more context.

unional:
https://www.npmjs.com/package/no-op
https://www.npmjs.com/package/node-noop
https://github.com/supershabam/nop
https://github.com/yoshuawuyts/noop2

All of these? Would be super easy to create anyway.... 😛

PR #522, opened 2016-06-09
add common/noop.json

bigsan:
**Typings URL:** https://github.com/bigsan/common-noop

The noop module that prevents TypeScript from complaining about "Cannot find module".

See [typings/typings#110](https://github.com/typings/typings/issues/110#issuecomment-191064423) and [typings/registry#164](https://github.com/typings/registry/issues/164) for more detail.

### DIFF
--- a/common/noop.json
+++ b/common/noop.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.1.0": "github:bigsan/common-noop#cb2c513367bf936da336eaf21a31ae4cd0d2ec1c"
+  }
+}


blakeembrey:
Do you think it's better to maybe call this `common~any`? Just trying to think about it, `noop` usually entails a function `noop`.

bigsan:
@blakeembrey Ya, the name `common~any` is more meaningful, I'll make new repos for that.

And, before sending new pull request, I think I should change the export style from `export = ...` to `export default ...`, it's more es6 syntax, do  you have any suggestion about that?

unional:
> I'll make new repos for that.

You can just rename it

blakeembrey:
@bigsan As it's a polyfill, I think you need to keep it as `export =`. Otherwise installing it as a filler for a CommonJS module won't work because they need to be imported as `import x = require('x')`.

PR #528, opened 2016-06-09
Add typings for meteor-rest

twastvedt:
**Typings URL:** https://github.com/twastvedt/typed-meteor-rest

* [X] Does the README explain the purpose of the typings and have a link to the JavaScript project? _Not sure what this means. What is the purpose of any typings beyond providing type definitions for a library?_
* [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)? _I'm also not sure how to answer this. I have an index.d.ts. The source structure isn't standard because it's a meteor package._

Add typings for the meteor atmosphere package simple:rest.


twastvedt:
Also, the typings generator created a blank `version` field in `typings.json`, so I filled that in with the version of meteor-rest that I wrote the typings against. I'm now not sure that was correct...

unional:
> Also, the typings generator created a blank version field in typings.json

Do you want to try out the latest version of `generator-typings`? In the latest version it reads the version from the npm/bower package directly (or ask you if it is not a npm/bower package).

unional:
> I filled that in with the version of meteor-rest

That is correct. 👍

unional:
> I'm also not sure how to answer this. I have an index.d.ts. The source structure isn't standard because it's a meteor package

Thanks for paying attention. :smile: I think this is fine.

twastvedt:
> Do you want to try out the latest version of generator-typings?
Cool, I will check that out next time. Thanks!

unional:
Just release `generator-typings@1.0.3` that comes from `update-notifier`. So from now on you will be notified if there is a new version available.

blakeembrey:
@twastvedt Can you move this to `global/`. I'll look at adding an atmosphere source soon. Quick question, how is atmosphere structured? Should Atmosphere in this repo be structured as `<namespace>/<pkg>.json`?

twastvedt:
@blakeembrey 

> Can you move this to global/
Oh right, sorry. Done.

> how is atmosphere structured?

I'm not entirely sure what you mean here, which maybe means that the namespace structure doesn't apply. Some packages add functionality to an existing namespace, as is the case with this package (meteor-rest). However some might do both that and also add a new global variable. I don't have an example at the moment, but I don't think it's required that a package export a single variable. Atmosphere is a hold-over from the days before modules.

Or am I missing something?

blakeembrey:
Mostly because I looked at https://atmospherejs.com/simple/rest and it seemed to be `simple/rest` as the name (though it replaced the colon with slash). Is this typical of atmosphere? Do all packages have a namespace (`simple`) and a package (`rest`) or is it an unlimited number of colons or no colons possible?

twastvedt:
Ah, I see. The part before the colon is the creator, like `twastvedt/typed-meteor-rest` on github. I believe that is the extent of its use.

PR #532, opened 2016-06-10
Add typings for meteor-collection-hooks

twastvedt:
**Typings URL:** https://github.com/twastvedt/typed-meteor-collection-hooks


twastvedt:
I forgot to add the .json extension to the new file. Trying to figure out how to make that change, as I created the pull request on github without first forking.

twastvedt:
Just realized this is in the wrong scope. Changing...

twastvedt:
I just can't win. Looking into this, I'll be back.

blakeembrey:
@twastvedt Thanks! Could you perhaps move it to the `global/` folder? I'll be happy to merge it then. Also, would it help to add a `meteor` folder? I'm not super familiar with the meteor ecosystem, but it seems they have their own package manager - are those packages global or CommonJS?

twastvedt:
@blakeembrey Ok. I guess I was confused about the difference between `global` and `env`. I saw meteor in `env`, so I put it there. 

As of 1.3 (3/28/2016) Meteor uses npm packages as well as their own Atmosphere packages. Atmosphere packages, at least all of those that I have used, are global. Currently Atmosphere packages afford a better integration with the Meteor environment. According to [the guide](http://guide.meteor.com/atmosphere-vs-npm.html) though, Atmosphere support will eventually be removed. I don't know when 'eventually' is, so I'm not sure whether it makes sense to add a source.

blakeembrey:
Thanks for the documentation. It's ok to add a source for meteor packages, even though it's legacy. It doesn't hurt us any to do that. If you'd like that, we can add a new `meteor` (or maybe `atmosphere`?) directory so people can add typings. Either way, I'm happy to merge the current PR for making it global.

There's some docs on structure in https://github.com/typings/registry#structure, but I understand it's a little iffy. The distinction between some things are a little arbitrary, but it makes it easier to recognise later when everything in `env` is a CLI tool or something similar (E.g. `node`) while everything in `global` is a library that just happens to be in JavaScript's global scope.

twastvedt:
An `atmosphere` source makes sense to me, but I feel unqualified to make that call!

> everything in env is a CLI tool or something similar (E.g. node) while everything in global is a library that just happens to be in JavaScript's global scope.

Aha. That makes much more sense. Thanks!

PR #534, opened 2016-06-10
Add typings for npm/tsconfig-glob

GiedriusGrabauskas:
**Typings URL:** https://github.com/QuatroTypings/npm-tsconfig-glob


blakeembrey:
This module is also already written in TypeScript. If they publish the `.d.ts` files, there's no need to write a separate definition for it.

blakeembrey:
Also, it's usually better to write things in the format I've written out below because that way users can still access the interfaces you're using for various reasons.

```ts
declare function main (options: main.Options): void;

declare namespace main {
  export interface Options {}
}

export = main;
```

PR #537, opened 2016-06-10
Add xhr-mock typings

effervescentia:
**Typings URL:** https://github.com/jameslnewell/xhr-mock


blakeembrey:
LGTM 👍 You can also change all those `declare`s to `export`s instead of having to use `export {}`.

PR #539, opened 2016-06-10
update export format of xhr-mock

effervescentia:
**Typings URL:** https://github.com/jameslnewell/xhr-mock

Updated the export format to export functions directly and only use the alias style for `delete` method


blakeembrey:
You should also be able to say `export function delete ...`.

effervescentia:
It claims its a reserved keyword and whatnot, although I can't say for
certain that I've tried the actual compiler, may have just been my linter.

blakeembrey:
You're right, thanks! 👍

PR #543, opened 2016-06-11
add common/any.json

bigsan:
**Typings URL:** https://github.com/bigsan/common-any

The common~any module.


blakeembrey:
Thanks! 👍

blakeembrey:
@bigsan I didn't notice that you had changed the export between this PR and the previous one. Would you be willing to change it back? See https://github.com/typings/registry/pull/522#issuecomment-225373990.

For example, you can't do the example below because the only property that exists after your change is `default`:

```ts
import co = require('co')

co(...)
```

PR #544, opened 2016-06-12
Update atom.json

unional:
https://github.com/typed-typings/env-atom

Rework the whole typings


unional:
@blakeembrey , I'm creating it as an module and use global augmentation to put `atom` on global.

It there a way to get around the build script?

unional:
Merging as original is not working anyway. 😛

PR #545, opened 2016-06-13
add chai-spies to registry

whenther:
**Typings URL:** https://github.com/whenther/typed-chai-spies

I took a crack at typing the chai-spies library, but ran into a couple problems:
- The library adds a `chai.spy()` function to the core `chai` object - but in trying to add that to the type definition, I'm running into the `Module augmentation cannot introduce new names into the top level scope.` error from [this issue](https://github.com/Microsoft/TypeScript/issues/6722). That's allowed on TS 1.9, so this works on `typescript@next`, but not on the stable channel. Is there a workaround that I don't know about?
- I used the yeoman generator, and I  think I followed the directions from it's [upgrade docs](https://github.com/typings/generator-typings/blob/master/UPGRADE.md) - but I must have something misconfigured, because I'm getting `cannot find module` errors in `test/test.ts`, even though when I import my `index.d.ts` into the project I'm using chai-spies in, everything is a-okay.
- Once I get the `cannot find module` issue sorted, I'd like to write some tests - but what's the difference between `test` and `source-test`?

Thanks!

### DIFF
--- a/npm/chai-spies.json
+++ b/npm/chai-spies.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "0.7.1": "github:whenther/typed-chai-spies#e8c648250e2fab1e1af54b0bb684ccd44d19cab8"
+    }
+}
--- a/npm/chai-spies.json
+++ b/npm/chai-spies.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.7.1": "github:whenther/typed-chai-spies#e8c648250e2fab1e1af54b0bb684ccd44d19cab8"
+        "0.7.1": "github:whenther/typed-chai-spies#d5dbba153d3995984edf74cf1562b9eb8436841b"
     }
 }


unional:
> because I'm getting cannot find module errors in test/test.ts

That's because you removed `node`. `blue-tape` has a global dependencies of `node`.
Fix it by `typings i -DG env~node`

`test` is where you write your test files. `source-test` is there if you can and want to utilize the tests from the source, so that you don't have to write your own tests (given the source test is good enough to cover the API).

In this case, if you want to do that, copy `source/test/spies.js` as `source-test/spies.ts` and configure the project to use `mocha` to run those tests.
It is currently a manual process and you can check out https://github.com/typed-typings/npm-batch as an example.

unional:
btw, I seriously need to rework `npm-chai` to the interface exposure. Didn't have time to. Would you like to help?

unional:
If it is fixed, I suspect you can just `declare module 'chai' {` and augment away.

whenther:
@unional thanks for the quick response! On the module errors - it turns out the problem was that I managed to forget to add the actual test file to the test directory's tsconfig. D'oh!

And cool, the source-tests make a lot of sense then! I ported the (nicely extensive) tests over from chai-spies, which found a bunch of missing stuff in my definitions, so that was helpful.

As for fixing the npm-chai definition - sure, I should have time to take a crack at it tomorrow. Hopefully I can find a way of switching this https://github.com/typed-typings/npm-chai/blob/master/lib/Chai.d.ts#L10 to an interface without breaking the other chai plugin definitions in the registry...

unional:
Those `***Static` were inherited from DT. I don't really like it and it should be do away.
Here is a nice example for re-export:
https://github.com/typed-typings/env-atom/blob/master/src/atom/atom.d.ts

Can do the same on `chai`

unional:
I'm going to merge this PR for now. If we can clean up `chai`, then we can just adjust this again.

Also, `chai` team is open to take the definitions into the package once it is ready. :smile:

PR #533, opened 2016-06-10
Add typings for npm/tsconfig-lint

GiedriusGrabauskas:
**Typings URL:** https://github.com/QuatroTypings/npm-tsconfig-lint

**Questions (for new typings):**

* [x] Does the README explain the purpose of the typings and have a link to the JavaScript project?
* [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
* [x] Are they external or global modules according the source (e.g. see README sources)?

PR #531, opened 2016-06-10
Add typings for npm/react-notification-system

GiedriusGrabauskas:
https://github.com/typings/registry/issues/498

**Typings URL:** https://github.com/QuatroTypings/npm-react-notification-system

**Questions (for new typings):**

* [x] Does the README explain the purpose of the typings and have a link to the JavaScript project?
* [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
* [x] Are they external or global modules according the source (e.g. see README sources)?

PR #520, opened 2016-06-08
Add typings for npm/gulp 3.9.1 and 4.0.0-alpha.2

GiedriusGrabauskas:
**Typings URL:** https://github.com/QuatroTypings/npm-gulp

**Questions (for new typings):**
* [X] Does the README explain the purpose of the typings and have a link to the JavaScript project?
* [X] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
* [X] Are they external or global modules according the source (e.g. see README sources)?

Created new repository with:
* Travis tests.
* Typings tests.
* Different versions.

PR #547, opened 2016-06-13
Packages structure fixes

GiedriusGrabauskas:
**Typings url:** https://github.com/QuatroTypings/npm-react-notification-system
**Fixes for:** https://github.com/typings/registry/pull/531

**Typings url:** https://github.com/QuatroTypings/npm-tsconfig-lint
**Fixes for:** https://github.com/typings/registry/pull/533

**Typings url:** https://github.com/QuatroTypings/npm-tsconfig-glob
**Fixes for:** https://github.com/typings/registry/pull/534

### DIFF
--- a/npm/npm-react-notification-system.json
+++ b/npm/npm-react-notification-system.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.2.7": "github:QuatroTypings/npm-react-notification-system/0.2.7#fd9f7e5b2a67dd76a5a762a9bf62b47cb1533973"
+    "0.2.7": "github:QuatroTypings/npm-react-notification-system/0.2.7#deadaa35cd0b677a6e3d8e3e7bdd558a107f3677"
   }
 }
--- a/npm/tsconfig-glob.json
+++ b/npm/tsconfig-glob.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.4.3": "github:QuatroTypings/npm-tsconfig-glob/0.4.3#ac08e4e9f967079c72ee89891316fef50f5ad54d"
+    "0.4.3": "github:QuatroTypings/npm-tsconfig-glob/0.4.3#802f1237c303039eaa06b9c1098a3386025423ef"
   }
 }
--- a/npm/tsconfig-lint.json
+++ b/npm/tsconfig-lint.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.11.0": "github:QuatroTypings/npm-tsconfig-lint/0.11.0#ca5903a97e151ad697165e19ee494fdff7de429b"
+    "0.11.0": "github:QuatroTypings/npm-tsconfig-lint/0.11.0#a1a441510a6d9a0b2ecc21920531b7258f291742"
   }
 }


blakeembrey:
Cheers! 👍 The only other thing I'd consider looking into is how you structure your module and exports. For example: https://github.com/QuatroTypings/npm-tsconfig-lint/blob/master/0.11.0/index.d.ts. Currently you export one member of the namespace, but it's a really inefficient way to write it because no one can access the rest of the module. You might want to use a pattern like: https://github.com/typings/typings/blob/master/docs/examples.md#function--utility. For example:

```ts
declare function tsconfigLint (options: tsconfigLint.Options, errorCallback: tsconfigLint.CallBack): void;

declare namespace tsconfigLint {
    interface Options {
        /**
         * A relative path from cwd to the directory containing a tsconfig.json. If not specified, the '.' is used.
         */
        configPath?: string;

        /**
         * The current working directory, defaults to `process.cwd()`
         */
        cwd?: string;

        /**
         * Whether or not `filesGlob` should be used in place of `files` for determining the files to lint.
         */
        useGlob?: boolean;

        tsconfigOptions?: {
            /**
             * The number of spaces to indent the tsconfig.json
             */
            indent?: number;
        };

        /**
         * A relative path from the configPath to the tslint configuration file.
         */
        tsLintConfigFilePath?: string;
    }

    interface CallBack {
        (err?: Error): void;
    }
}

export = tsconfigLint;
```

With this pattern, people can still use `import x = require('...'); x.Options`.

DovydasNavickas:
What is not nice about the format you've written in https://github.com/typings/registry/pull/534 is that code completion gets polluted with all of the interfaces.

image

We we're looking at this format with @GiedriusGrabauskas, but we thought that the one with `contracts.d.ts` would be a bit cleaner one.

image

And if I understand correctly, what you are saying in https://github.com/typings/registry/pull/520
>No one can import anything from your file this way and that file (`contracts`) does not actually exist.

You are only partly right.

It is true that the file does not exist in the package itself.
But as there are ONLY interfaces in the `contracts.d.ts`, it is not required to exist in the package. Because TypeScript compiler emits `import` (`require`) statements in the compiled code ONLY if something that requires emitting is used from the import.

And in case of module defined in `contracts.d.ts`, it has ONLY interfaces and does not require emitting, which is why TypeScript leaves its import out when compiling.

You can try to compile `test.ts` and see for yourself that there is only `import * as gulp from 'gulp';` emitted and nothing is emitted for line `import * as GulpClient from 'gulp/contracts';`

blakeembrey:
Yes, you're correct - but no one knows to do that and there's no documentation on it.

blakeembrey:
On top of which, if you _really_ wanted to hide the interfaces, you don't need a namespace for that.

GiedriusGrabauskas:
I agree with you, that I should use `declare function` instead of `declare var`.
> ...because no one can access the rest of the module.

I added working sample in test.
[QuatroTypings/npm-tsconfig-lint/test.ts](https://github.com/QuatroTypings/npm-tsconfig-lint/blob/master/0.11.0/test/test.ts)

blakeembrey:
@GiedriusGrabauskas Cheers! Sorry to cause a philosophical debate on this 😄 

I really don't see a big issue with the pollution here. If you had written it in TypeScript, it would be the same, and I think we should be aiming to write it as it would be written in TypeScript - not some hacks to make it "prettier" when really the ordering behaviour could just be changed here to put interfaces toward the bottom. Also, in this module format, the interfaces are only exposed on `x.*` - so it's not pollution the main export of a function.

unional:
@GiedriusGrabauskas I had the same reason, I also wanted to keep the completion clean and simple. But in the end I give in to this statement:

> we should be aiming to write it as it would be written in TypeScript

🌷

DovydasNavickas:
Hopefully there is gonna be more and more things written in TypeScript and we are not gonna have these kind of discussions :clap:

PR #550, opened 2016-06-14
add aurelia-http-utils 0.1.2

heruan:
**Typings URL:** 

### DIFF
--- a/npm/aurelia-http-utils.json
+++ b/npm/aurelia-http-utils.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.1.2": "github:heruan/aurelia-http-utils#8bcd82893227dd1b4616780b83548d88cc9b1bdb"
+  }
+}


unional:
Since you are the package author, you probably don't need to add it to the registry. 🌷

heruan:
Thank you @unional for the feedback. I'm trying to figure out what is needed and what is not for definitions to be referenced in a package I build! After publishing the package, I installed it in my project with JSPM and on build the compiler complained about missing definitions; I then tried

```shell
$ typings install --save aurelia-http-utils
typings ERR! message Unable to find "aurelia-http-utils" ("npm") in the registry. Did you want to try searching another source? Also, if you want contribute these typings, please help us: https://github.com/typings/registry
typings ERR! caused by https://api.typings.org/entries/npm/aurelia-http-utils/versions/latest responded with 404, expected it to equal 200
```

and so I guessed I had to put it in the registry.

unional:
Jspm is not natively supported by `tsc`. I'm working on a `jspm:` source so you can do `typings I jspm:<pkg>`

heruan:
Thank you @unional I'll look forward to it. Still, I can add the typings with the `github:` source, if course; but I'd like to have them on the default registry as well for a simpler installation procedure. Isn't this the same for official Aurelia's packages? They have their own typings, but they were also published on the registry.

If there is any kind of policy to prevent adding to the registry packages which already defines their own typings, I will follow it—I'm just curious 🌻

blakeembrey:
@heruan The policy is to do the best thing for all TypeScript consumers possible. If the typings can exist in the package, they should be there first. For that reason, we'll also remove packages from the registry if they get merged into the packages themselves. For installing without the registry, you can use a scheme:

```
typings install --save npm:aurelia-http-utils
```

However, if there's no dependencies on typings from Typings, you can natively use the `typings` field in `package.json` and skip using Typings altogether.

heruan:
Thank you @blakeembrey I had read the contributing guidelines, I was just a bit confused since I've seen Aurelia typings published in the registry, referring `typed-typings` repositories instead of being bundled in their own packages themselves, but I understand it was just temporary. Thank you!

unional:
Ah, I see your confusion. I added those typings for `aurelia` they are needed because the one in the package itself is in the wrong format. The Aurelia team knows about that and they are working on a solution. Once they fix it, these repos can be removed.

unional:
Just check, seems like `aurelia` have fixed their typings for RC! I'll validate if those added to the repo are all RC (probably should be), then I can deprecate those in the registry. :smile:

PR #552, opened 2016-06-14
Update version for react-redux

yeerkkiller1:
react-redux version update.

### DIFF
--- a/npm/react-redux.json
+++ b/npm/react-redux.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "4.4.0": "github:andrew-w-ross/typings-react-redux#6be88a57427f8ec220cad13d68565fd5f5711146"
+    "4.4.0": "github:andrew-w-ross/typings-react-redux#21202533f75a73d4fa4c50e0357aaf23739fcabb"
   }
 }


unional:
Looks good. One thing you can improve is:
https://github.com/andrew-w-ross/typings-react-redux/blob/edb8cdf9483f041d3e31c94ee5e4b310b65468fb/react-redux.d.ts#L25-L27
```ts
	function wrapWithConnect<T extends ComponentConstructor<any>>(
		component: T
	): T

// to
interface WrapBlah<T extends ComponentConstructor<any>> {
    (component: T) => T
}
```
then you can skip the `typeof`

Above is sudo code and is just an idea. 🌷

yeerkkiller1:
Thanks, I was just updating the typings reference and didn't look too closely at them. 

I tried your suggestion though, and I might be doing something wrong, but I think typeof does something really weird to make the generic type resolve at a later stage.

I tried it: [http://www.typescriptlang.org/play/index.html#src=interface%20iFnc%3CT%3E%7B(v%3A%20T)%3A%20T%7D%3B%0D%0Adeclare%20function%20wrapperFnc%3CT%3E()%3A%20iFnc%3CT%3E%3B%0D%0Avar%20returned%20%3D%20wrapperFnc()(%7Bx%3A%205%7D)%3B%0D%0Areturned.x%20%3D%202%3B%0D%0A%0D%0Adeclare%20function%20fnc%3CT%3E(v%3A%20T)%3A%20T%3B%0D%0Adeclare%20function%20wrapperFnc2%3CT%3E()%3A%20typeof%20fnc%3B%0D%0Avar%20returned2%20%3D%20wrapperFnc2()(%7Bx%3A%205%7D)%3B%0D%0Areturned2.x%20%3D%202%3B]

```ts
interface iFnc<T>{(v: T): T};
declare function wrapperFnc<T>(): iFnc<T>;
var returned = wrapperFnc()({x: 5});
returned.x = 2;

declare function fnc<T>(v: T): T;
declare function wrapperFnc2<T>(): typeof fnc;
var returned2 = wrapperFnc2()({x: 5});
returned2.x = 2;
```

But when the typeof isn't used it doesn't infer T.

It is definitely very strange.

unional:
Yeah, that's strange. Maybe I have missed something obvious? Do you want to post this to SO to see what did we miss or it is a bug on TS?

unional:
Ok, this works:

```ts
interface iFnc{ <T>(v: T): T};
declare function wrapperFnc<T>(): iFnc;
var returned = wrapperFnc()({x: 5});
returned.x = 2;
```

So the problem is why does `interface iFnc<T> { (v: T): T }` does not work.

PR #553, opened 2016-06-15
Add typings for npm/gulp 3.9.1 and 4.0.0-alpha.2

GiedriusGrabauskas:
**Typings URL:** https://github.com/typed-typings/npm-gulp

Old pull request: #520

### DIFF
--- a/npm/gulp.json
+++ b/npm/gulp.json
@@ -1,5 +1,7 @@
 {
   "versions": {
-    "3.8.0": "github:typed-typings/npm-gulp#504da08753acdfe9dd2c415ab8fe36151e9645c1"
+    "3.8.0": "github:typed-typings/npm-gulp/3.8.0#4ed9a4a3275559c73a396eff7e1fde3824951ebb",
+    "3.9.1": "github:typed-typings/npm-gulp/3.9.1#4ed9a4a3275559c73a396eff7e1fde3824951ebb",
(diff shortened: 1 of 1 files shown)


blakeembrey:
If possible, we need to fix these dependencies ASAP: https://github.com/typed-typings/npm-gulp/blob/master/4.0.0-alpha.2/typings.json#L6-L15. It's not useful to have things in the registry broken.

blakeembrey:
If you also get a chance, please fix the exports: https://github.com/typed-typings/npm-gulp/blob/master/4.0.0-alpha.2/index.d.ts#L206. Currently nothing other than that export can be accessed, making the namespace redundant.

DovydasNavickas:
Isn't it that identical name namespace `GulpClient` and declared var `GulpClient` kind of "extend" each other? In code completion when importing gulp as:
```
import * as gulp from 'gulp';
```
and writing:
```
gulp.
```
we can see both things: interfaces from the namespace and the functions from interface `Gulp`:

image

blakeembrey:
Yes, you're right. It's a very non-idiomatic way to do it. This whole thing should just be top-level exports without any need for a namespace, let alone this namespace interface trick.

blakeembrey:
For example, it could just be:

```ts
export function src(...): ...;
```

PR #563, opened 2016-06-16
Update aurelia-dependency-injection.json

unional:
Remove
https://github.com/typed-typings/github-aurelia-dependency-injection
and aurelia-pal

### DIFF
--- a/npm/aurelia-dependency-injection.json
+++ b/npm/aurelia-dependency-injection.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0-beta.1.2.0": "github:typed-typings/github-aurelia-dependency-injection#16600ce1463c3f99f3e9d2fa3b2b2458ef174fc7"
+    "1.0.0-beta.1.2.0": "github:typed-typings/github-aurelia-dependency-injection#187e0639488c70580f78f8ffe004bce872815264"
   }
 }
--- a/npm/aurelia-dependency-injection.json
+++ b/npm/aurelia-dependency-injection.json
@@ -1,5 +1,0 @@
-{
-  "versions": {
-    "1.0.0-beta.1.2.0": "github:typed-typings/github-aurelia-dependency-injection#187e0639488c70580f78f8ffe004bce872815264"
-  }
-}
--- a/npm/aurelia-pal.json
+++ b/npm/aurelia-pal.json
@@ -1,6 +1,0 @@
-{
-  "versions": {
-    "1.0.0-beta.1.2.0": "github:typed-typings/npm-aurelia-pal#b75df0b87f5a150243a6fa02bb1c5f30521fa026",
-    "1.0.0-beta.1.2.2": "github:typed-typings/npm-aurelia-pal#fe2a8c625bf60482c143ee039f866c49a8b03f90"
-  }
-}


blakeembrey:
@unional Why not delete it instead? That'd stop it showing up in search but will continue being searchable. The other possible solution might be to add a deprecated message to the registry entries themselves since we won't always have access to the original source - thoughts?

unional:
Was thinking about a graceful way to do it. If I just delete it, then when consumer `typings install` it would error out saying XYZ does not exists.

On the other hand, I notice the `postmessage` is not working. In process of filing an issue. 😛

unional:
> The other possible solution might be to add a deprecated message to the registry entries themselves

What / how to do that?

blakeembrey:
@unional Anyone can install if they have the reference. E.g. it's in their `typings.json` file. So `typings install` definitely stays working, but `typings install aurelia-dependency-injection` will say it doesn't exist.

blakeembrey:
Also, we could change the message for `typings install ...` errors to say it no longer exists in the registry if they _were_ previously versions but none are currently available.

blakeembrey:
> What / how to do that?

Not currently support, just brainstorming - I suspect we add a field alongside `"versions"` or inside versions to mark as a deprecation message?

unional:
> Not currently support, just brainstorming - I suspect we add a field alongside "versions" or inside versions to mark as a deprecation message?

Or the install will get the current `postmessage` to display it? (or add a `deprecated` flag or `deprecatedMessage`)



> but typings install aurelia-dependency-injection will say it doesn't exist.

Or say "it is deprecated" if we can tell registry about it, somehow.

unional:
One problem is we can't let existing user know that the typings is deprecated (as you said `typings install` will remain working). But maybe that's a minor issue.

blakeembrey:
Well, that's why I used to log when it's been removed from the registry/updated. Could add that back again? When the entry is removed, log that it's been removed during `typings install`?

> Or say "it is deprecated" if we can tell registry about it, somehow.

I feel like the best way to tell the registry is to delete it. We can update the 404 to instead return something else when all versions in the registry are deprecated.

unional:
> We can update the 404 to instead return something else when all versions in the registry are deprecated.

Don't like the idea of hunting the log to determine if 404 means not found or deprecated. Could run into bigger performance issue in the long run.
Also, when deprecated, a version is still needed, because the typings is added to the source package starting from certain version. If consumer using a previous version, they still and should be able to install it through `typings install aurelia-dependency-injection`

blakeembrey:
You know the API itself has a copy of all versions ever published right? Well, now you do. There's no hunting about, the API itself never hits GitHub (that would destroy performance, wouldn't be able to handle the current load).

blakeembrey:
Also, that's what I just said - that still works when you delete it.

blakeembrey:
I wouldn't build a system that randomly breaks people's projects 😝

blakeembrey:
For example, no moment from the registry in search (http://api.typings.org/search?name=moment), the entry is a 404 (http://api.typings.org/entries/npm/moment), versions is a 404 (http://api.typings.org/entries/npm/moment/versions), but you can find deprecated (http://api.typings.org/entries/npm/moment/versions?deprecated=true), and you can **always** access previous versions (http://api.typings.org/entries/npm/moment/tags/2.10.5+20160211003958 <- this is what the client hits during `typings install` - always resolves a tag deprecated or not) or search for a particular version also works (http://api.typings.org/entries/npm/moment/versions/%3E=2)

PR #564, opened 2016-06-16
new hashtag add return void on compose function

jnath:


unional:
Thank.
I would be best if you include your typings repo link in the description. 🌷

PR #573, opened 2016-06-21
type definitions for google-libphonenumber

mrshaq:
**Typings URL:** https://github.com/mrshaq/typed-google-libphonenumber

Added basic type definitions for google-libphonenumber.


blakeembrey:
Thanks. I'll trust you on the definition, since I honestly have no idea when it comes to that ecosystem. Why define the module name (https://github.com/mrshaq/typed-google-libphonenumber/blob/master/node.d.ts#L22-L24)?

mrshaq:
It's not a complete definition. Just the essential parts, although I'll try to improve it over time.
I was under the impression that since it's an NPM module I need to define the module name. Is there a better way to do it?

blakeembrey:
If you're typing the NPM definition, it should be https://github.com/seegno/google-libphonenumber written in external module format and placed in the `npm/` folder (not `global`).

PR #585, opened 2016-06-23
Update versions

barbatus:
https://github.com/meteor-typings/meteor

Update to the versions with last minor corrections.

Commit with changes 
https://github.com/meteor-typings/meteor/commit/6a5a5df634213aec1ead932e2c1c0a29b500c0af

### DIFF
--- a/env/meteor.json
+++ b/env/meteor.json
@@ -1,6 +1,5 @@
 {
   "versions": {
-    "1.3.0": "github:meteor-typings/meteor/1.3#64f0d62cb163049fd2c04c71ac1a8402b4ff6c9e",
     "1.2.0": "github:meteor-typings/meteor/1.2#64f0d62cb163049fd2c04c71ac1a8402b4ff6c9e"
   }
 }
--- a/env/meteor.json
+++ b/env/meteor.json
@@ -1,5 +1,6 @@
 {
   "versions": {
-    "1.3.0": "github:barbatus/meteor-typings/meteor/1.3#83c6933d0ef07abfd4a2a85b5ec41bc3949f6d28"
+    "1.3.0": "github:meteor-typings/meteor/1.3#64f0d62cb163049fd2c04c71ac1a8402b4ff6c9e",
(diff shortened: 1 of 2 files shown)


blakeembrey:
Thanks! I won't go through it all, but here (https://github.com/meteor-typings/meteor/commit/6a5a5df634213aec1ead932e2c1c0a29b500c0af#diff-1fb387bd43e1e058a776173ee2b5ef97L275) you can use string literals which might help. E.g. `type Status = 'connected' | 'connecting' | 'failed' | 'waiting' | 'offline'`.

blakeembrey:
By the way, would you like access to @typed-typings to move your repositories there?

barbatus:
I've tried `type Status = 'connected' | 'connecting' | 'failed' | 'waiting' | 'offline'`, it's cursing `error TS1110: Type expected`.

Not sure about moving repos to @typed-typings. I wanted to keep them together to let people look up them easily (some of those libs can't be used outside Meteor anyways)

blakeembrey:
What `tsc` version are you using? Sounds like < 1.8 since that's when string literals were introduced.

On @typed-typings, you're welcome to move them if you want. I'm just worried about future maintenance of some packages, I've recently had to fork a couple there. However, that's not a big deal overall (GitHub supports it nicely), just figure people can collaborate better with more people involved.

barbatus:
@blakeembrey you're right I got some old typescript installed locally. thanks for the tip!

I plan to add more people to @meteor-typings. Hope future maintenance won't be an issue.

barbatus:
@blakeembrey Do you mind a question regarding updating from 0.x to 1.0. Sorry didn't find better place.
Docs say that in order to control typings resolution user needs to setup `resolution: {main: .., browser: ...}`. That's ok, but is any way to setup default resolution? For example, https://github.com/meteor-typings/meteor/blob/master/1.3/typings.json. I tested in this way but it seems to place everything into the globals folder.

blakeembrey:
@barbatus You might need to expand on it, but resolution isn't related to global vs non-global - it's related to following the main vs browser paths. It's also not something you configure as an author, but as a consumer.

PR #591, opened 2016-06-28
fuse.js ES6 Typings

kurtaqui:
**Typings URL: https://github.com/kurtaqui/npm-fuse.js

Add typings from definitely typed which supports ES6 imports.

### DIFF
--- a/npm/fuse.json
+++ b/npm/fuse.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.0": "github.com:kurtaqui/npm-fuse.js#db27865d0327004a2fa2bc8254fa1590aa1d3f27"
+    "0.1.0": "github.com:kurtaqui/npm-fuse.js#8fffdee3581116964b5f4d272315a14ffba05a2b"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
Typings under NPM should be written in external module format - don't use `declare module "..."` (https://github.com/kurtaqui/npm-fuse.js/blob/master/fuse.js.d.ts#L25-L28).

blakeembrey:
Here's some correct (and improved) typings:

```ts
declare class Fuse {
	constructor (list: any[], options?: Fuse.Options);
	search<T>(pattern: string): T[];
}

declare namespace Fuse {
	export interface Options {
		id?: string;
		caseSensitive?: boolean;
		include?: string[];
		shouldSort?: boolean;
		searchFn?: any;
		sortFn?: (a: { score: number }, b: { score: number }) => number;
		getFn?: (obj: any, path: string) => any;
		keys?: string[] | { name: string; weight: number }[];
		verbose?: boolean;
		tokenize?: boolean;
		location?: number;
		distance?: number;
		threshold?: number;
		maxPatternLength?: number;
	}
}

export = Fuse;
```

blakeembrey:
Thanks, it's looking better. The existing issue is that you've written `github.com:` instead of `github:`.

kurtaqui:
Thanks have missed that, and thanks for the help on the typings :)

Issue #597, opened 2016-07-03
Typings for soundcloud


shawnlauzon:
I'm working on it right now, suggestions are much appreciated when I submit pull request.

API: https://developers.soundcloud.com/docs/api/sdks#http

PR #598, opened 2016-07-03
Add Soundcloud typings

shawnlauzon:
See Issue #597

**Typings URL:** shawnlauzon/typed-soundcloud


unional:
Most likely there is a better way to expose it. The current construct is generally discouraged because user or anyone who want to use or extends the types can't access the types directly.

You can consider exposing methods inside `SoundCloud` directly.

unional:
Also, you should use `es6-promise` instead of relying on the global `promise`

PR #599, opened 2016-07-03
Add typings for Vuex v1.0.0-rc.2

ktsn:
**Typings URL:** https://github.com/vuejs/vuex/tree/master/types

  * The typings is in the JavaScript project

### DIFF
--- a/global/vuex.json
+++ b/global/vuex.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0-rc": "github:vuejs/vuex/types#645bb9c46d5ca5bc3c991a35107d6142080e1e7f"
+    "1.0.0-rc.2": "github:vuejs/vuex/types#eca13a5b8223296ff0eebfd7efd9bd9509e364a8"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
@ktsn If you're the author of Vuex, you shouldn't be adding the module to the registry. The typings should be provided as the `typings` field in `package.json` instead - this would allow them to be versioned together, correctly.

blakeembrey:
The module should also be written in external module definition format. I would recommend you get in touch with the module author and try to make it more useful to consumers. In the current form, it's really just more work than it's worth to maintain.

ktsn:
@blakeembrey Thank you for your feedback.
The reason that I wrote the module in `declare module` format is Vuex has `import logger from 'vuex/logger'` import style defined on [this](https://github.com/vuejs/vuex/blob/master/types/index.d.ts#L91)

How do I handle this import style in external module definition format?

blakeembrey:
You can write two definition files and store them alongside the JavaScript files. It would work just the same as though it were originally written in TypeScript.

ktsn:
Oh, I see. I will do that.
Thank you very much for your support!

PR #606, opened 2016-07-06
Update Freewall definitions

Maddosaurus:
**Typings URL:** https://github.com/Maddosaurus/typed-freewall

I had to update the Freewall type definitions because there was a previously undocumented function in there.

Thanks!


unional:
Seems like you are on the wrong branch thus there is merge conflict. Can to fix that or create a new PR?

PR #608, opened 2016-07-06
Delete redux-thunk.json

unional:
`redux-thunk` now supplies its own typings


unional:
Not released yet. Need to wait for post 2.1.0. DO NOT MERGE

PR #620, opened 2016-07-12
Added reference to sebastianhaas/typed-jszip.

sebastianhaas:
**Typings URL:** https://github.com/sebastianhaas/typed-jszip

* [ ] Are the tests running

I took most of the definitions from DefinitelyTyped, as stated in the readme.

There are still two problems existing, I don't really know how to resolve them so I would be glad if somebody could help out.

***Definitions issue***
JSZip uses a constructor-like function to create an instance of `JSZip`. Since they both have the same name, but the first one is an interface in a namespace, using it in typescript will currently result in a rather ugly 
```typescript
  let zip: JSZip.JSZip = new JSZip();
```
Is there a way to make JSZip both the name of an interface as well as an object holding a constructor like function?
https://stuk.github.io/jszip/documentation/api_jszip/constructor.html

***Browser test issue***
 I wrote tests for all major use-cases, server-side runs fine, but I don't really have experience with `jspm` and I can't get it running on the browser-side. It appears that generator-typings is not producing ready-to-use framework for the browser atm.


blakeembrey:
For your issue, I'd recommend you just do a refactor of the definition into a single class and re-export it. It would make for nicer intellisense also. For instance:

```ts
// Make the main export a class.
declare class JSZip {
  constructor ();
  constructor (data: any, options?: JSZip.Options);
}

// Use the namespace to store interfaces and such.
declare namespace {
  export interface Options {
    ...
  }
}

// Export `JSZip` `module.exports =` style.
export = JSZip
```

blakeembrey:
Oh, actually - I see you might also want to support the plain function (non-class) style of constructor. I can't remember where I've done that style before, but you can't use the class syntax for it. I'll think on it, I feel I've done that someplace before and should be possible - I might submit a PR later if I try it out.

felixfbecker:
@blakeembrey class syntax supports calling without `new`? Not afaik. If you find it out please hit me up because I'm using the static/instance interface pattern everywhere and it bloats definition files.

blakeembrey:
@felixfbecker You can't, I mentioned it above (https://github.com/typings/registry/pull/620#issuecomment-232081157) - I'd just overlooked that usage. The preferred way to do it would be an inline value/type. Something like:

```
interface JSZip {
  something: boolean
}

var JSZip: {
  (): JSZip
  new (): JSZip
}
```

felixfbecker:
@blakeembrey I misread your comment, I thought you said you _can_ use the class syntax for it 😅

PR #624, opened 2016-07-12
add vsts-task-lib typing

bryanmacfarlane:
**Typings URL:** https://github.com/Microsoft/typed-vsts-task-lib

Initial typing.  The typing is generated as part of the build from: https://github.com/Microsoft/vsts-task-lib

The lib has been in production for Visual Studio Team System build tasks for awhile.  But I'm new to typings (started today), so forgive rookie mistakes.

### DIFF
--- a/npm/vsts-task-lib.json
+++ b/npm/vsts-task-lib.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.8.4": "github:Microsoft/typed-vsts-task-lib#c2adcd15e82efdb8085e626c14dbe837cc9e7a2b"
+        "0.8.4": "github:Microsoft/typed-vsts-task-lib#2fdfa0a0234439627ba28eb7c04e66ba4f9cc4eb"
     }
 }
--- a/npm/vsts-task-lib.json
+++ b/npm/vsts-task-lib.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.8.4": "github:Microsoft/typed-vsts-task-lib#65585b30e352b2abae2465f95b059c2e0033d3d5"
+        "0.8.4": "github:Microsoft/typed-vsts-task-lib#f38f18c5823b533925d0ed3d9b5cb91652fd32f1"
     }
 }
--- a/npm/vsts-task-lib.json
+++ b/npm/vsts-task-lib.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "0.8.4": "https://github.com/Microsoft/typed-vsts-task-lib#65585b30e352b2abae2465f95b059c2e0033d3d5"
+    }
+}
--- a/npm/vsts-task-lib.json
+++ b/npm/vsts-task-lib.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.8.4": "github:Microsoft/typed-vsts-task-lib#f38f18c5823b533925d0ed3d9b5cb91652fd32f1"
+        "0.8.4": "github:Microsoft/typed-vsts-task-lib#c2adcd15e82efdb8085e626c14dbe837cc9e7a2b"
     }
 }
--- a/npm/vsts-task-lib.json
+++ b/npm/vsts-task-lib.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.8.4": "github:Microsoft/typed-vsts-task-lib#31271b6079c174e734e927a9d82a8495ddea59af"
+        "0.9.0": "github:Microsoft/typed-vsts-task-lib#3f17c01c52a13c4c2ba0701082f2759636769697"
     }
 }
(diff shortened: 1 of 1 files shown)


unional:
Is there any reason you don't want to add the typings directly in  https://github.com/Microsoft/vsts-task-lib?

unional:
It currently fails because it is a global script, not external module. You need to unwrap the `declare module 'abc'` and use `declare namespace ...` or export them at top level.

bryanmacfarlane:
@unional - I could do that.  I read we were supposed to create a repo called typed-[name]

The issue I'm having right now, is it works locally (incuding tests)

```bash
$ git clone https://github.com/Microsoft/typed-vsts-task-lib test-task-lib
$ cd test-task-lib/
$ npm install

> typed-vsts-task-lib@0.8.4 postinstall /Users/bryan/Projects/test-task-lib
> typings install && cd test && typings install --global


├── node (global)
└── q (global)


├── node (global)
├── q (global)
└── vsts-task-lib (global)

$ npm test

> typed-vsts-task-lib@0.8.4 test /Users/bryan/Projects/test-task-lib
> tsc -p test

$ echo $?
0
```

But when run by travis in PR, it fails:

https://travis-ci.org/typings/registry/builds/144305036

unional:
That's because in your test code you are treating it as global, thus it works.
https://github.com/Microsoft/typed-vsts-task-lib/blob/master/test/typings.json#L8

However, it should really be an external module, not global, as you use it through `import tl = require('vsts-task-lib/task');`

unional:
https://github.com/typings/typings/blob/master/docs/external-modules.md
https://github.com/typings/typings/blob/master/docs/faq.md#types-of-typings

bryanmacfarlane:
@unional - thanks for help so far.

My build uses dts-generator as part of the build and produces _build/.d.ts/vsts-task-lib.d.ts.  It's unclear to me how to export at the top level. 

Consumers of the lib already do import tl = require('vsts-task-lib/task'); so I want to retain that as well.

Thanks again.

bryanmacfarlane:
I think the dts-generator was causing my issues - trying to use the separate files outputted by tsc declaration true and going that route

bryanmacfarlane:
Yes, that was it.  For generation I was using something to bundle after my build (dts-generator but there's dts-bundle).  I changed to having compiler do it (tsc --declaration true) which generated multiple files and I used those instead.

blakeembrey:
@bryanmacfarlane If you're the module author, it would be easier to maintain the definition in your own repo instead. There's some guidance in the README (https://github.com/typings/registry#as-package-author) but let me know if there's anything else we can help with.

bryanmacfarlane:
Makes sense.

After reading through the links my take away is since I release on npm, ship the js and .d.ts files with the npm package and point to the .d.ts entry with typings property of package.json.  then, in the typings registry, I point to npm:{nameInNpmRegistry}

I'll make those changes.  Please ack if that sounds right.

blakeembrey:
@bryanmacfarlane Correct, but you don't need to update the Typings registry at all then. It'll just work, then you can avoid the registry altogether.

bryanmacfarlane:
Ahh.  Awesome

bryanmacfarlane:
Hopefully last question ...

I do above with typings entry in package.json and publish
From the link it seems like typescript will resolve types by looking in node_modules and everything works for consumers that just install my package.  

But, they need to also pull typings for types from dependencies I expose (q, node).  If I published in my own typed-xxx repo the typings.json would declare the dependences

So do users do `typings install npm~vsts-task-lib` to get it to pull down node and q typings?

Does that mean I should have a typings.json in the package in addition to the package.json property?

blakeembrey:
> But, they need to also pull typings for types from dependencies I expose (q, node). If I published in my own typed-xxx repo the typings.json would declare the dependences

Correct, yes. But you can also do that from your published module on NPM also. However, global sub-dependencies are never installed by Typings (only printed that the users needs to install them).

> So do users do typings install npm~vsts-task-lib to get it to pull down node and q typings?

Yes.

> Does that mean I should have a typings.json in the package in addition to the package.json property?

You can. That's what I do currently, for example in https://github.com/blakeembrey/popsicle. Then someone needs to install it separately using `typings install npm:popsicle`. It's not ideal, but the process is going to be getting better with TypeScript 2.0.

unional:
Yes, you should keep the `typings.json` in there.
Use would need to do `typings install -save npm:vsts-task-lib` to install the typings locally (from `node_modules/...`. notice it is `:`, not `~`).

However, in your case it will not really install `q` and `node` because both of them are global.
`typings install` will print out message saying so and user need to manually install them.

There is no other way for `node`. For `q`, if the typings were in module format (published under `typings/registry/npm/q.json`), then `typings install` will bundle and install it automatically.

bryanmacfarlane:
Thanks everyone.  I got it all working.

Typescript (and especially new async await) is a game changer for node.  Keep up the good work.

Closing this PR since I'm distributed with my npm package.

PR #627, opened 2016-07-13
Add declaration for bignumber.js

felixfbecker:
**Typings URL:** https://github.com/felixfbecker/typed-bignumber.js

* [x] Does the README explain the purpose of the typings and have a link to the JavaScript project? _yes_
* [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)? _yes_
* [x] Are they external or global modules according the source (e.g. see README sources)? _external_

### DIFF
--- a/npm/bignumber.js.json
+++ b/npm/bignumber.js.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.3.0": "github:felixfbecker/typed-bignumber.js#d93c7452a6d8a11c9f01b2c410a7b13cd6859927"
+  }
+}


blakeembrey:
That's some amazing documentation!

The only thing I'd note is that you _may_ want to put the interfaces into a `BigNumber` namespace. This would allow people to still access those interfaces outside the type definition, which can be important sometimes. E.g.

```ts
const config: BigNumber.Configuation = { ... }
```

felixfbecker:
@blakeembrey Thanks :)
I opened https://github.com/MikeMcl/bignumber.js/issues/97 to hopefully be able to remote `export =` completely, which would allow to export the interfaces. I think namespace / interface merging only works with interfaces, not classes - don't really want to declare a global namespace.

blakeembrey:
No, it does work. It's not global either.

```ts
declare class BigNumber {}
declare namespace BigNumber {}

export = BigNumber;
```

blakeembrey:
Here's an example of it working to a different project I contributed to: https://github.com/kevinbeaty/any-promise/blob/master/index.d.ts.

felixfbecker:
@blakeembrey So then I have to reference every interface like this, right?
```ts
declare namespace BigNumber {
  export interface Configuration {}
}
declare class BigNumber {
  config(config: BigNumber.Configuration): BigNumber.Configuration;
}
```
I would like to avoid that :/

blakeembrey:
It's nicer than not allowing anyone to access the interfaces.

unional:
I should update the example to reflect this. :)

felixfbecker:
I will remove `export = ` instead thanks to https://github.com/MikeMcl/bignumber.js/commit/34b79d1a0cbca866469cf75abdc62aecd40dd35d :)

PR #632, opened 2016-07-17
feat: Added co to global

flauc:
**Typings URL:** https://github.com/flauc/typed-co

**Questions:**


blakeembrey:
I don't think this is the right place to add `co`. Can you please write it as an external module for NPM? Also, the definition could be greatly improved. In it's current form I'm not willing to accept it, as it's pretty much just `any` for everything. It should be typed to support the correct function type (generators) and extract the correct return type in generators. It should also return a promise for `wrap`.

PR #644, opened 2016-07-22
Create ip-address.json

westonpace:
**Typings URL:** https://github.com/westonpace/typed-ip-address

### DIFF
--- a/npm/ip-address.json
+++ b/npm/ip-address.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "5.8.0": "github:westonpace/typed-ip-address#5dbea66d1e5b2d176d3a996e112493452d467157"
+    "5.8.0": "github:westonpace/typed-ip-address#44d5d32ad89a9c4ca98fc6e58984cd082730a0f9"
   }
 }
--- a/npm/ip-address.json
+++ b/npm/ip-address.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "5.8.0": "github:westonpace/typed-ip-address#a29f324842868a4f7a1fc83694303924e377cc0e"
+    "5.8.0": "github:westonpace/typed-ip-address#58a76a00bbed18afb50a10bfe823f4539f5cd1ea"
   }
 }
--- a/npm/ip-address.json
+++ b/npm/ip-address.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "5.8.0": "github:westonpace/typed-ip-address#44d5d32ad89a9c4ca98fc6e58984cd082730a0f9"
+    "5.8.0": "github:westonpace/typed-ip-address#a29f324842868a4f7a1fc83694303924e377cc0e"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
Why not just write it in normal ES6-style instead of exporting a namespace?

```ts
export class Address6 {}
```

Also, can you make the classes that aren't classes namespaces instead? (E.g. https://github.com/westonpace/typed-ip-address/blob/master/ip-address.d.ts#L3-L54). It's mostly a matter of renaming it, but it really helps people use the library properly - right now, if I see it's a class and tried to use it I'd run into type errors. Using an interface is what you actually want.

unional:
It's part of `generator-typings` fault. When you choose to create a `commonjs` typings, it is defaulted to:

```ts
declare namespace abc {}
export = abc;
```

If you chose `ES2015`, then it would be

```ts
export default {}
```

westonpace:
That seems likely.  When I was generating the project I wasn't as clear in understanding the distinction between `commonjs` and `ES2015` as I now am so I probably meant to choose `ES2015`.

As for the classes that aren't classes did you mean I should make them namespaces or interfaces?  I am new to typescript and don't really understand what makes them not classes.

I am copying the pattern I see in Angular2, e.g. https://github.com/angular/angular/blob/2.0.0-rc.4/modules/%40angular/http/src/interfaces.ts#L48-L55 (admittedly I should have used interface)

blakeembrey:
It looks great, don't get me wrong 😄 Just wanted to sort out the couple of kinks. So, for writing type definitions, either works (namespace vs ES6 module-style) since they are just definitions for the JavaScript source. ES6-style is usually the "more natural" way of writing it, but I'm happy to support the other format also.

For making them interfaces, it's pretty much just replacing the word `class` with `interface` in your definition. They shouldn't be namespaces though, just interfaces. The difference is that a namespace is kind of like an object you can add to later in JavaScript, while a class is an object (right, it's an instance of an object) that has a constructor function, while interfaces are purely type annotations - using the former two in a value position would not error, while using an interface as a value will. If you want to share `Address` members, you might want to use an abstract class in TypeScript (so no one accidentally imports it and uses it).

Let me know if this helps! If you're not sure where to start, just say so. I'm happy to submit a PR getting most of the way there together.

westonpace:
Ok, removed namespace, changed to interfaces and made Address abstract.  Please take another look and thanks for all your input and help.

unional:
There is a page in TypeScript handbook with a table the clearly describe the various types. but I can't find it right now.

unional:
https://www.typescriptlang.org/docs/handbook/declaration-merging.html

blakeembrey:
Nice! You don't need the final `export default` though - that doesn't exist in the library (https://github.com/westonpace/typed-ip-address/blob/master/ip-address.d.ts#L118).

westonpace:
Ok, I think I got it this time.

blakeembrey:
Thanks! It looks great 😄

PR #646, opened 2016-07-24
Create find-up.json

unional:
https://github.com/types/npm-find-up

@blakeembrey in `find-up` it uses `pinkie-promise` and I follow it so:
https://github.com/types/npm-find-up/blob/master/index.d.ts#L1

Should I be using `any-promise` or `es6-promise` or just `PromiseLike` instead?

### DIFF
--- a/npm/find-up.json
+++ b/npm/find-up.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.1.2": "github:types/npm-find-up#33fc4719e4433245f7467f349dd1ebd812231d28"
+  }
+}


blakeembrey:
Using `pinkie-promise` looks good to me. I propose using the correct `Promise` libraries for return types, but `PromiseLike` for input types - that would be maximumly compatible.

Should we make a doc in @typings for these sorts of decisions? It's the sort of thing we should definitely share with people writing new typings. I feel like we've decided at least a dozen things like this, TypeScript put together a good initial doc for doing things like using `void`, etc.

unional:
> Should we make a doc in @typings for these sorts of decisions?

👍 👍 👍 👍 👍 
And can I put them into https://github.com/unional/typescript-guidelines also?

PR #647, opened 2016-07-24
Delete any-promise.json

unional:
`any-promise` now carries its own typings.

But this could break some builds.
Should we do it?


blakeembrey:
Yews, it's fine. Remember old references still hang around, so it's ok :)

PR #648, opened 2016-07-25
add definition for mailgen

vothanhkiet:
**Typings URL:** https://github.com/vothanhkiet/mailgen-typings
**Source URL:** https://github.com/eladnava/mailgen

* [x] Does the README explain the purpose of the typings and have a link to the JavaScript project?  - Yes
* [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)? - Yes
* [x] Are they external or global modules according the source (e.g. see README sources)? - External

### DIFF
--- a/npm/mailgen.json
+++ b/npm/mailgen.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.25": "github:vothanhkiet/mailgen-typings#acb2e07083ebc9d6955d21045531cd99aea38397"
+    "1.0.25": "github:vothanhkiet/mailgen-typings#fe51953d856447fb2fa020721129fc01ede99d26"
   }
 }
--- a/npm/mailgen.json
+++ b/npm/mailgen.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.25": "github:vothanhkiet/mailgen-typings#cd362124aebe2b2515f059aeac71f8ff66250a65"
+    "1.0.25": "github:vothanhkiet/mailgen-typings#acb2e07083ebc9d6955d21045531cd99aea38397"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
Sorry for the delay. Looking at the module compared to your typings, it looks like there's a bug. If you look at https://github.com/eladnava/mailgen#usage, the usage is `export =` style and not `export Mailgen`. Can you refactor to make sure it follows the correct export styles?

vothanhkiet:
You're right. I've just refactor, remove all export and only use 

> export = Mailgen 

However, when I use in my code, tsc throw error ```TS2497: Module ''mailgen'' resolves to a non-module entity and cannot be imported using this construct.```. The compile code still work, I received the email . After add 

> namespace Mailgen {
> }

I don't get error anymore. Can you explain this ? I'm still new at TypeScript

P/S: I will make new pull request again.

blakeembrey:
You don't need to make a new PR, just update this one.

The way to import a CommonJS module is `import x = require('x')`. Using `import * as x from 'x'` is a fallback for ES6-modules, but TypeScript also made the decision to support namespaces for that style also since all modules pre-1.5 were written like that (1.5 was the introduction of ES6 support). You can read more here: https://www.typescriptlang.org/docs/handbook/modules.html.

blakeembrey:
You might want to move the interfaces (https://github.com/vothanhkiet/mailgen-typings/blob/master/index.d.ts#L4-L35) into the namespace so you can access them externally to the module.

vothanhkiet:
Thanks for pointing that out, can you recheck again please ?

blakeembrey:
Never seen that approach (https://github.com/vothanhkiet/mailgen-typings/blob/master/index.d.ts#L1-L2). If it works, it works 😄

PR #651, opened 2016-07-28
Added the version 1.10.1 of oracledb to typings registry.

Bigous:
**Typings URL:** https://github.com/Bigous/typed-oracledb

### DIFF
--- a/npm/oracledb.json
+++ b/npm/oracledb.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.10.1": "github:Bigous/typed-oracledb#6d9cb41096084ac3ce67ba6f45d20a737728a155"
+    "1.10.1": "github:Bigous/typed-oracledb#d9946b4fd482afc4e8b2bd41788a5e49704ad7c5"
   }
 }
--- a/npm/oracledb.json
+++ b/npm/oracledb.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.10.1": "github:Bigous/typed-oracledb#6d9cb41096084ac3ce67ba6f45d20a737728a155"
+  }
+}


unional:
Two things I spotted:

- You are using `stream`, so should list `node` as `globalDependencies` instead of `globalDevDependencies`
- Since you are simply exporting the whole namespace, you can export what's inside the namespace directly with ES2015 named export syntax.

Bigous:
Hi @unional ,
First point you are correct! Done.
The second one I didn't get. Why is it better export each item on namespace than export the whole namespace (I'm doing it in DefinitelyTyped, but I thought that exporting the namespace would be more interesting - less typing)?

unional:
Second point allows user to import using both commonjs and ES2015 in a more consistent way, relying on ES2015 -> CommonJS interop.

The other way relies on CommonJS -> ES2015 interop is more problematic (due to differences between `SystemJS` and `Babel`)

It is a minor difference, but the ES2015 named export is the preferred approach, in general.

Of course, that would be a debate on whether the typings should stay true to the source. So the counter argument is whether if the source is written in ES2015 or not.

The final choice would be up to you. 🌷

unional:
cc/ @blakeembrey if there is something I missed or made incorrect statements 🌷

blakeembrey:
Looks correct 👍 In the end it makes little difference. You do that on DefinitelyTyped because you have to - there's no external modules (in the past), so you'd make a namespace to expose things globally and allow others to extend it. It's not super important and I think it's good either way, we just tend to promote using ES6-style `export function foo` instead of `namespace X { export function foo } export = X` as it's more natural and generally how the original author intended it.

Bigous:
Just for the record - the generator-typings (which I used here to make things happen faster) creates a index.d.ts with the following content:
```typescript
declare module xxxx {
}
export = xxxx;
```
It induced me to write that way.

But now I've got the point - next version I'll use the ES6 approach.

unional:
Yup. That's the template for commonjs modules. Because they can have function + namespace structure.

blakeembrey:
No worries. Thanks for your contribution 😄 I'll stop pointing it out from now on, I'm happy with either style.

PR #655, opened 2016-07-30
update riot.json

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-riot

**Change Summary:**

- fix `mixin` declarations
- add global `mixin` declaration
- update `root` type to be custom extension of `HTMLElement` instead of `Node`

### DIFF
--- a/npm/riot.json
+++ b/npm/riot.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "2.3.0": "github:effervescentia/typed-riot#db2af9ea9d2ab08d3e56313e0c91e3a5f050d0cc"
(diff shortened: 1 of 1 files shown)


unional:
You need to reorder the overloads. The first one will always win.

unional:
https://github.com/unional/typescript-guidelines/blob/master/pages/typings/overloading.md

blakeembrey:
@unional Although correct, in this case because all the arguments are required (none optional) it should always resolve to the correct one (based on arity). I agree keeping it in mind is important though, and making the order more clear can help (and help you not forget this fact).

unional:
Oh, you are correct. 👍

PR #656, opened 2016-07-31
Add `servicestack-client` to registry.

Layoric:
**Typings URL:** https://github.com/ServiceStack/servicestack-client/blob/master/typings.json

GitHub repository for NPM module with JS. Native TypeScript library, wanting to share d.ts files via both to allow flexibility of use. 

Yes, in ./lib folder

External as it is available on NPM.

Let me know if any other info is required :+1:

### DIFF
--- a/npm/servicestack-client.json
+++ b/npm/servicestack-client.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.0.9": "github:ServiceStack/servicestack-client#0dea064e98505ba5f1b70450ccf1b1ce16fdde4d"
+    "0.0.10": "github:ServiceStack/servicestack-client#1a3c40c74581682efc3345c73d8a2ec038de46fc"
   }
 }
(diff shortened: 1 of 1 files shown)


unional:
As the error suggest, you need to write it in external module format.
You can see some examples here:
https://github.com/typings/typings/blob/master/docs/examples.md
Remove the `declare module '...' {`.

Layoric:
@unional Thanks for the info. 

I think I might have misunderstood the different "sources". 

Given this is a http client library and (to me at least) makes the most sense to import and use it as such:
`import {JsonServiceClient} from 'servicestack-client';`
Might this make more sense as a global source rather than removing the `declare module`?

Thanks again.

unional:
It should be kept as external module format. It's a module, not a global script.

You don't need to use `declare module 'servicestack-client' {` to tell it is named `servicestack-client`. It is defined in the registry (`npm/servicestack-client.json`) and can be changed by the consumer.

See if this would make sense to you: https://github.com/typings/typings/blob/master/docs/external-modules.md

unional:
👍

PR #661, opened 2016-08-02
Update bluebird.json

unional:
for https://github.com/types/npm-bluebird/pull/23


blakeembrey:
Left a comment about the predicate object. It should be purely an index signature.

PR #664, opened 2016-08-02
Add 'preact' typings to global

kruczy:
**Typings URL:** https://github.com/developit/preact/blob/master/src/preact.d.ts

the typings are in the project source

yes, `preact.js` -> `preact.d.ts`

**global**

### DIFF
--- a/global/preact.json
+++ b/global/preact.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "5.6.0": "github:developit/preact#c1bc415fdc04e84a245d6c688393f40134f3ac68"
+  }
+}


blakeembrey:
As the author, it would probably be easier to provide instructions on the README for TypeScript users instead of relying on the registry and a third-party update process that may be out-of-sync. What do you think?

kruczy:
the typings could be referenced in `package.json` `typings` property, the problem here is that these need to be global in 1.8.10 because a global JSX namespace needs to be defined and augmenting global scope from modules does not work there

The situation is different in 2.0 as you can augment global scope in modules but if you add the  `typings` property and anyone with ts 1.8 uses the lib they will get a compile error

That's why I think for 1.8 users a `typings install global~preact --global --save` should be fine
and for ts2 users `typings install preact --save` (ill make a pr for this later unless there is a better way to do it)

of course it would be better to just have a `package.json` entry and have a seamless experiance, though I don't think it is possible in the current state of things (unless I am missing something)

blakeembrey:
It's not just a `typings` field in `package.json` though - you can just provide direction to do `typings install npm:preact/src/preact.d.ts --global` after you've NPM installed. You can also distribute a `typings.json` with `global: true` if needed that will work with `typings install npm:preact`. In this case, you can probably skip Typings altogether by recommending they add it to their project context - E.g. `files` in `tsconfig.json` or using a reference.

kruczy:
hey, could do that `typings install npm:preact` seems fine for now, when TS2 goes stable we could probably put the `typings` prop in `package.json`

blakeembrey:
@kruczy Why can't you just do a `typings` property today then? Nothing there changes with 2.0.

kruczy:
you cannot augment global scope in 1.8 from the module, it will give a compilation error if you try
and the global JSX namespace needs to be augmented to make this work,

another way to do it could be to just expose the types and have the user define the JSX namespace

PR #667, opened 2016-08-03
Update react-router.json

unional:
https://github.com/types/npm-react-router/pull/5
https://github.com/types/npm-react-router/pull/4

Update version
Update typings.json to use global instead of ambient.
Fix https://github.com/types/npm-react-router/issues/3


blakeembrey:
Thank you 😄

PR #669, opened 2016-08-08
add typings for filter-object

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-filter-object


blakeembrey:
FWIW, you can do 😄 

```ts
declare function filterObject(obj: any, glob: string): any;
declare function filterObject(obj: any, globs: string[]): any;

export = filterObject;
```

PR #670, opened 2016-08-09
Add typings for stripe

taxpon:
**Typings URL:** https://github.com/taxpon/typings-stripe

### DIFF
--- a/global/stripe.json
+++ b/global/stripe.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:taxpon/typings-stripe#829cc194419b9d3d860e4551e106a4335a2da2aa"
+  }
+}
--- a/global/stripe.json
+++ b/global/stripe.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:taxpon/typings-stripe#829cc194419b9d3d860e4551e106a4335a2da2aa"
+    "1.0.0": "github:taxpon/typings-stripe#cf87a3d2141543bd4db26cf5c567329961ebec09"
   }
 }


blakeembrey:
Not sure if I'm missing something, but the first example uses `require('stripe')('token')` - these typings won't be able to handle that?

amooz:
@taxpon any progress answering @blakeembrey's questions?  I was about to create new definitions for stripe but found your waiting pull request first.

taxpon:
@blakeembrey @amooz Sorry for late response. I just created type definition which is correspond to [this in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/stripe/stripe.d.ts). I updated some interfaces. 

@blakeembrey 
> Not sure if I'm missing something, but the first example uses require('stripe')('token') - these typings won't be able to handle that?

This is another definition, maybe [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7171d1588156a7c5702f11a37bd4490e86f08665/stripe-node/stripe-node.d.ts). So not included in this PR.

blakeembrey:
@taxpon Can you link the docs/library this is for then? In the README it says it's https://stripe.com/docs/api, and those are the ones I looked at.

taxpon:
@blakeembrey Sorry for unclear document. I fixed README in [typings-stripe](https://github.com/taxpon/typings-stripe) repository. Please check it.

blakeembrey:
@taxpon I'm still confused what these typings are for though. Can you share the project that this types?

@amooz Were you after the node.js typings?

taxpon:
@blakeembrey I can not share you my private repo, but this typing helps developers to build [payment form with stripe.js](https://stripe.com/docs/custom-form).

blakeembrey:
@taxpon Thanks. I don't need your private repo, that's ok 😄 I just want to figure out what the definitions are actually for. That link helps - so it's the `stripe.js` library. Is it the `v2` implementation too? If it's only a global, you shouldn't need to declare a module called `Stripe` to use it (https://github.com/taxpon/typings-stripe/blob/master/stripe.d.ts#L116-L119).

Can you link to those `stripe.js` docs? It's helpful (https://stripe.com/docs/custom-form).

taxpon:
@blakeembrey I think this typings support v2 and I updated [README](https://github.com/taxpon/typings-stripe), so it would be great if you review it :)

blakeembrey:
@taxpon Looks good. Would you like to update this PR? You've got a small typo on the readme (thier -> their) and https://github.com/taxpon/typings-stripe/blob/master/typings.json#L4 should be `"global": true` (for 1.0).

taxpon:
@blakeembrey Oops.. I fixed it. Please review.

blakeembrey:
Thanks @taxpon!

PR #681, opened 2016-08-12
Jacada Interact Extension

doronguttman:
**Typings URL:** https://github.com/jacadainc/typed-interact-extension

The README explains the purpose, but the JavaScript is proprietary and not available to link to.
 
There is specifically associated js file.

yes

### DIFF
--- a/global/jacada-interact-extension.json
+++ b/global/jacada-interact-extension.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "7.5.8": "github:jacadainc/typed-interact-extension#4b2cb082cc66ca7a98213252499be15bb0284124"
+    "7.5.8": "github:jacadainc/typed-interact-extension#52fa94f5499a6c8a6aca3ae987f8cb88dff48c15"
   }
 }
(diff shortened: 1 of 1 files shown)


blakeembrey:
@doronguttman Thanks for the contribution. I don't know a lot about how you distribute the definition, but if it's proprietary and you're the author would it make more sense to provide the definition/instructions with the library itself?

doronguttman:
You may be right, though, the library is a 3rd-party, which we have no control over, but would like to provide users a simple method to install typings when they develop extensions.

blakeembrey:
Ok, I've merged it now 👍

doronguttman:
Thanks!

doronguttman:
@blakeembrey I just made an update to the typings repo and the hash I used in the PR is no longer valid.
This brings out a question. Is there a way not to define a specific hash on the typings so you always get the latest commit?

unional:
It is not advised to do that because that means whenever you make a commit you can potentially break all of your consumers.
Think about someone running travis and suddenly it starts failing without any code change.

doronguttman:
I see your point, but how is that different from me updating the repo, then doing a PR to update the typings repo?

unional:
Because the user won't get any intermittent commits. When a PR is filed here, we can help taking a look to make sure the typings is in good shape.

unional:
Also, user won't automatically gets the updated typings unless they do `typings install <pkg>`
`typings install` would download and use the same old commit.

PR #689, opened 2016-08-13
fixed TypingsError on switchery

beenotung:
**Typings URL:** 
https://github.com/StefanKoenen/npm-switchery

moved switchery to global as suggested by the npm script ```test-all```


blakeembrey:
I think it's just an old commit that was obscure - it looks like it was updated since then, would you want to update to `077593a06deb0243de2f27a92133c8d1e0e50ec4` instead?

blakeembrey:
Closed with https://github.com/typings/registry/commit/22677250689b1ae3f3d6fcbd82db40798a134d39.

PR #690, opened 2016-08-13
Delete es6-promise.json

unional:
now integrated to source.

### DIFF
--- a/npm/es6-promise.json
+++ b/npm/es6-promise.json
@@ -1,5 +1,0 @@
-{
-  "versions": {
-    "3.0.0": "github:types/npm-es6-promise#fb04188767acfec1defd054fc8024fafa5cd4de7"
-  }
-}


felixfbecker:
What about targets like Node 5 that have promises but cannot use es6 lib.d.ts?

unional:
this is for `es6-promise`. It is integrated to the source repo. Will be release very soon. Already merged
https://github.com/stefanpenner/es6-promise/pull/211

felixfbecker:
Ah, okay. But then we should at least wait until the release is out.

unional:
Yes, thus `waiting for response` tag.

blakeembrey:
I'm still a little concerned about people using this definition that isn't particularly the `es6-promise` module. I'm guilty of using it like this 😄 Perhaps we should move this to `common/`?

PR #694, opened 2016-08-15
Added support for regex and arrays  in Express.Router.use

IOAyman:
**Typings URL:** https://github.com/ioayman/typed-express

Added support for regex and arrays in Express.Router.use
`Express.Router.use( path: RexExp | string | (RegExp | string)[], ....`

reference: http://expressjs.com/en/4x/api.html#app.use


felixfbecker:
You should open a PR at my repository, not set the reference to your typings.

IOAyman:
Ok.

Issue #53, opened 2016-01-22
Add check for the references to test suite


blakeembrey:
References should use commit hash and not master. This should be part of the test suite.

PR #706, opened 2016-08-17
Add check for the references to test suite

pellejacobs:
> References should use commit hash and not master. This should be part of the test suite.

#53


blakeembrey:
@pellejacobs Looks like a good change. Let me export `parseDependency` from the root so you're not deep requiring into the module.

blakeembrey:
Thanks!

PR #709, opened 2016-08-19
Create mem-fs-editor

unional:
https://github.com/types/mem-fs-editor


unional:
Seems like a bug on either travis (more likely) or github. Close and re-create

PR #714, opened 2016-08-21
Create html2json.json

unional:
https://github.com/types/html2json

### DIFF
--- a/npm/html2json.json
+++ b/npm/html2json.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.2": "github:types/html2json#572182333d86673af294eb1da98f1750bcaefd68"
+  }
+}


blakeembrey:
@unional Can we stick with the existing naming scheme until we resolve what to do? 😄

unional:
Anyway to push and get it resolved? If there is a chance to have `types/*` as source of `@types/*`, naming it straight is probably the way to go.
The only question is what to do with non-npm packages, right?
Should not affect naming decision for npm packages, I guess.

blakeembrey:
@unional Push a decision? I don't think it's really needed, there's no relation to the GitHub repo and NPM names, so we don't really need to conflate those and should just stick to keeping things consistent for now. You can do `npm install types/npm-html2json` which installs the module named `@types/html2json`. I'd like to keep it consistent since we have various packages in there are may event have none types as we go (E.g. with https://github.com/types/guidelines).

unional:
True. Ok. 🌷

unional:
Speaking of which, maybe the current scheme can be proposed to type-publisher?

blakeembrey:
Which is the current scheme you refer to? Just curious for your take on what the proposal would be - e.g. renaming on NPM or GitHub?

unional:
As is. Github: `types/<source>-<pkg>`.
For npm package, `types/npm-<pkg>` maps to `@types/<pkg>` (and should not be `@types/npm-<pkg>` to avoid confusion in the npm ecosystem).

For other packages, we can do `types/env-<pkg>` maps to `@types/env-<pkg>` or just decide their scheme in the future.

blakeembrey:
Ah, ok. I think we just map it all onto `@types` for now and hope there's no conflict to mirror TS 2.0 + DefinitelyTyped. So that would be `@types/<pkg>` in both cases.

Issue #713, opened 2016-08-20
RFC: typings for phoenix


akeating:
https://github.com/akeating/typed-phoenix

I created typed definitions for phoenix framework and am considering contributing them to typings . 

The phoenix [lib](https://www.npmjs.com/package/phoenix) is written in es6. I rewrote it in Typescript and generated the definition (and a source) that is closely compatible with the original source. I plan to maintain this by modifying my Typescript source as Chris McCord's underlying lib changes.

Are there any obvious issues or changes I'll need to make to ensure compatibility with the Typings registry and ready the repo for a pr?

blakeembrey:
It looks ok, but I'd recommend using it with `noImplicitAny` as I think a few people won't be able to use the current definition if they have that enabled.

felixfbecker:
Rewriting the source in TS seems like a lot of effort 😲

akeating:
@blakeembrey yeah we're ok as I used the following command-line to generate the definition:
```
tsc lib/phoenix.ts -d --noImplicitAny --noImplicitReturns
```

akeating:
@felixfbecker I discovered hand-rolling the definition worked for minimal usage. Overall though the process seemed error-prone and time-consuming. I decided the tradeoff was worth it mainly because the  underlying lib was written in es6 which is not too far from Typescript with a little effort. The generation process gave me a fairly comprehensive baseline definition to improve, I learnt a lot about Typescript and I suspect the total time involved to be roughly equivalent.

felixfbecker:
@akeating It probably was a good idea to copy the source and start from there if it is ES6. But I think it will be a pain if you want to keep the source up-to-date. I would remove the implementations and rename the files to .d.ts, it will be easier for long-term maintenance.

akeating:
I suspect you're right about that. It should be easier to look for signature changes than diffing and refactoring the source.

akeating:
One question I do have though. I marked many properties as _private_ which had the consequence of removing the typing from the generated definition. What are the definition best practices when it comes to access modifiers? The definition does lose a lot of information by using _private_ but on the other hand access modifiers clarify the author's intention. This distinction is perhaps less important in javascript. Given the underlying source is javascript, I am inclined to remove the _private_ access modifiers so the resulting definition has a richer set of typings.

felixfbecker:
`private` means the user should not use it, typically what is prefixed with an underscore in JS. It can be beneficial to include it in the typings because it prevents a user from overriding a private property without noticing.

[...]

PR #715, opened 2016-08-21
Support for phoenix framework

akeating:
**Typings URL:** https://github.com/akeating/typed-phoenix

See also #713 .


felixfbecker:
https://github.com/akeating/typed-phoenix/blob/master/priv/static/phoenix.d.ts#L1

Instead of declaring and exporting a namespace, just directly export all the types

akeating:
Updated PTAL.

blakeembrey:
LGTM /cc @felixfbecker 

By the way, you can just type `export class` and `export interface`, you don't need to use that particular export syntax.

PR #716, opened 2016-08-23
rename repo names

unional:
for `simple-git`, `mem-fs`, `mem-fs-editor` and `html2json`


felixfbecker:
So is this decided now? Because _if_ we gonna rename repos here we should rename all of them or none.

unional:
I kept forgetting that the repo names and npm package name can be separated.
So for now, keep it the way it originally is.

PR #728, opened 2016-08-26
Merging typings-monk

sfali23:
**Typings URL:** https://github.com/sfali23/typings-monk

### DIFF
--- a/npm/monk.json
+++ b/npm/monk.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.1.2": "github:sfali23/typings-monk#2220fd7d31530e93b5dc1d390eb1f2b0e8fa34e0"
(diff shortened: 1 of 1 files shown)


felixfbecker:
https://github.com/sfali23/typings-monk/blob/master/test/monk-tests.ts#L5
Should be 
```ts
import monk = require('monk');
```
Otherwise it will just be typed `any`. You can prevent such errors by setting `noImplicitAny` in your tsconfig.json. You would then have to compile the declarations with `typings bundle`, and try to compile the test with `tsc`. If there are no compilation errors, the typings can be considered to be correct (and I am pretty sure they arent, see the feedback I gave you).

It always makes sense to copy examples from the official docs to the tests.

You can find more help here https://github.com/types/guidelines

sfali23:
Made required changes in deceleration file:

- bundle is getting generated by `typings build`
- library can be imported now by `import monk = require('monk');`

felixfbecker:
LGTM!

PR #731, opened 2016-08-30
linked typings for archieml

ejmurra:
**Typings URL:** https://github.com/ejmurra/typed-archieml

### DIFF
--- a/npm/archieml.json
+++ b/npm/archieml.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.4.2": "github:ejmurra/typed-archieml#a1e4739ffb57a51f7209a21d9b7868b639f76d27"
+    "0.4.2": "github:ejmurra/typed-archieml#800ff60fa346877438b612c03fb6b0f8216a8680"
   }
 }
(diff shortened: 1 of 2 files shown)


felixfbecker:
You don't need to `declare namespace`, just export the individual types directly like `export interface ...`

ejmurra:
Thanks for the pointer. I used the typings-generator and selected commonjs, amd, and script tags for how the package can be used and then I used the output that it generated directly. 

It seems there's a [pattern suggested](https://www.typescriptlang.org/docs/handbook/writing-declaration-files.html#global--external-agnostic-libraries) by the Typescript docs which declares a `namespace` and then declares a `module` which exports it.

Which is the preferred way of doing it? Should I just export the interface directly and the function on a `namespace`?

Sorry for the noobiness, this is my first time contributing a type definition. Thanks!

felixfbecker:
That is the example _Global / External-agnostic Libraries_, which means a library that both exports elements and introduces a global namespace. But typings you write for NPM modules _are_ already "external" modules (aka ES6 modules). They don't introduce anything global _unless_ you use a `declare global { ... }` block _or_ set `global: true` in your typings.json. That's why you don't need to use `declare module "package-name"` or a namespace (except in certain cases, but this isn't one).

ejmurra:
That makes sense, thank you!

felixfbecker:
Change the file to
```ts
export interface AmlOptions {
  comments: boolean;
}

export function load(content: string, opts?: AmlOptions): any;
```

PR #734, opened 2016-08-31
Added oracledb v1.11.0 to the registry

Bigous:
Added version 1.11.0
 -- Pool Cache
 -- getPool
 -- Database Types
 -- Connection and Pool became EventEmitter


felixfbecker:
I don't know if it's worth maintaining a list of minor versions, typically we only reference major versions. @blakeembrey ?

blakeembrey:
If those are literally additions in a minor, seems good to me 😄

PR #748, opened 2016-09-02
Update node.json

blakeembrey:
**Typings URL:** https://github.com/types/env-node

Tidied up a couple of types: https://github.com/types/env-node/commits/master.

### DIFF
--- a/env/node.json
+++ b/env/node.json
@@ -1,10 +1,10 @@
 {
   "versions": {
-    "0.8.0": "github:types/env-node/0.8#c581dc01fea68adffa2f290fcda20f7cb917f752",
-    "0.10.0": "github:types/env-node/0.10#c581dc01fea68adffa2f290fcda20f7cb917f752",
-    "0.11.0": "github:types/env-node/0.11#c581dc01fea68adffa2f290fcda20f7cb917f752",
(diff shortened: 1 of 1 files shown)


zelein:
This merge has broken all of my projects that are using the env-node typings. It has specifically broken references to the "express-serve-static-core" project as well as the "types-is" project.

blakeembrey:
@zelein Are these projects from DefinitelyTyped? I removed the deprecated methods because of exactly this reason, they kept being used and shouldn't be. Express has correct definitions under `npm`, and I can type `type-is` today if that is helpful.

zelein:
Typing out the type-is would be very helpful if you can spare the time for
it. We have an internal declaration file, but it is basic and out of date.

The more I'm using TS the more I'm seeing the typings declarations to be
much more up to date than their DT counterparts.

blakeembrey:
@zelein Sorry for the delay, it took me days to finish the RethinkDB typings (it's insane!). I quickly typed out `type-is`, you can see the definition here: https://github.com/types/npm-type-is. It's a pretty small one 😄

zelein:
Thank you @blakeembrey. Much appreciated!

PR #750, opened 2016-09-03
Create koa.json

jkeylu:
**Typings URL:** https://github.com/types/npm-koa

### DIFF
--- a/npm/koa.json
+++ b/npm/koa.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "2.0.0": "github:types/npm-koa#afc9613c0aaf8406b3afd95bd2f051e7d339b231"
+  }
+}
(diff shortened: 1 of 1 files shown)


felixfbecker:
Is it possible to use classes? Or will the class/namespace merging not work?
https://github.com/koajs/koa/blob/v2.x/lib/application.js#L32

felixfbecker:
https://github.com/types/npm-koa/blob/master/lib/application.d.ts#L65-L67
I think exporting type aliases to the interfaces would be better suited here

jkeylu:
You're right

2016-09-03 16:42 GMT+08:00 Felix Becker <notifications@github.com>:

felixfbecker:
LGTM

PR #751, opened 2016-09-04
create meteor-migrations.json

niceilm:
**Typings URL:** https://github.com/niceilm/typed-meteor-migrations

* [v] Does the README explain the purpose of the typings and have a link to the JavaScript project?
* [] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?  
no, the typing is in a single file
* [v] Are they external or global modules according the source (e.g. see README sources)?  
global module

### DIFF
--- a/global/meteor-migrations.json
+++ b/global/meteor-migrations.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:niceilm/typed-meteor-migrations#c2841241e9589545d35707c510b11497e3829759"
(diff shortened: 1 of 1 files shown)


blakeembrey:
Thanks! 

This pattern (https://github.com/niceilm/typed-meteor-migrations/blob/master/meteor-migrations.d.ts#L24-L30) can also be written using the inline syntax - `var Migrations: {}` - instead, in case you're interested 😄

PR #757, opened 2016-09-05
Create co.json

blakeembrey:
**Typings URL:** https://github.com/types/npm-co

If anyone is interested, it's a pretty decent demonstration of TypeScript generics and how to unravel a type (E.g using `<T> (arg: Promise<T> | T): Promise<T>` to unravel promises and other types - in this case, thunks, iterables, etc). However, the type definition itself isn't very effective because of the limitations on generator functions in TypeScript. It appears the return type isn't even considered. Also, yielded values can't get any types either - they are always `any` and the input to `IterableIterators` is also always `any`.

/cc @mhegazy am I using the generators correctly? Is it possible to use the return type **only** or otherwise specify the result of `yield`s?

### DIFF
--- a/npm/co.json
+++ b/npm/co.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "4.6.0": "github:types/npm-co#650c90a3f097bbc2f8bded21d7d8586e2eb8ac91"
+  }
+}


mhegazy:
Currently generators only capture the type of the yield expression. issue https://github.com/Microsoft/TypeScript/issues/2983 tracks handling it correctly and capturing the return expression type as well.

PR #768, opened 2016-09-11
Add typings for mock-knex

jessezhang91:
**Typings URL:** https://github.com/jessezhang91/typed-mock-knex

       - single index.d.ts
       - external modules

### DIFF
--- a/npm/mock-knex.json
+++ b/npm/mock-knex.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.3.0": "github:jessezhang91/typed-mock-knex#fd10876c268c0f711218753d1ec939e5ab60380c"
+  }
+}


felixfbecker:
https://github.com/jessezhang91/typed-mock-knex/blob/master/index.d.ts#L4
Please remove the `declare namespace` and export the symbols directly

felixfbecker:
https://github.com/jessezhang91/typed-mock-knex/blob/master/typings.json#L7-L8
These should not be `devDependencies`

It would be super awesome if you could convert `knex` too, so we get proper dependency management and users don't have to install it manually :)

jessezhang91:
Should those be global dependencies?

felixfbecker:
`node` yes. `knex` unfortunately too, because all typings from DT are "global" by nature. Which is why it would be nice to have `knex` in the registry aswell, because otherwise users have to install it manually.

felixfbecker:
https://github.com/jessezhang91/typed-mock-knex/blob/master/index.d.ts#L1
This should be `import knex = require('knex')`

jessezhang91:
Gotcha. I'll try to convert the `knex` typings over some time soon.

jessezhang91:
Alright, I've made the dependency and require changes. Thanks!

felixfbecker:
Nice :)

PR #787, opened 2016-09-18
Add types for darrylwest/mock-browser

nohira:
**Typings URL:** https://github.com/nohira/typed-mock-browser

How can I determine the last point?

### DIFF
--- a/global/mock-browser.json
+++ b/global/mock-browser.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.92.12": "github:nohira/typed-mock-browser#57d5083f5ed5c6f69d29b56d4b10fc002ed4829c"
+  }
+}
--- a/global/mock-browser.json
+++ b/global/mock-browser.json
@@ -1,5 +1,0 @@
-{
-  "versions": {
-    "0.92.12": "github:nohira/typed-mock-browser#57d5083f5ed5c6f69d29b56d4b10fc002ed4829c"
-  }
-}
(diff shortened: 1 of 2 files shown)


felixfbecker:
These typings are global and should go under global/ and have a global: true flag set

nohira:
I've moved to global.
Can you just explain how you determine if it's global or not?

blakeembrey:
@nohira the criteria is "does it have an `import` or `export` at the top-level". This is TypeScript's semantics, when it does it makes the definition a module and when it does not it makes it a global definition. More nuance, we try to write certain modules in either global or external module format depending on the consumption - it wouldn't make sense to write a global version of an NPM module, for instance. Does this help? 😄

nohira:
@blakeembrey Yeah, I guess
Do you mean it's more in the syntax and the usage?
That I can write more like in the [nock types](https://github.com/nohira/npm-nock/blob/master/index.d.ts)?

blakeembrey:
@nohira Correct, yeah. You'd want to write it in external module format and have an export at the top level (removing https://github.com/nohira/typed-mock-browser/blob/master/index.d.ts#L3-L5). Otherwise, it's considered "global" and people might not connect it to that NPM module.

nohira:
@blakeembrey I've changed to non global format. I hope I did understand well.
Please tell me if something is wrong.

blakeembrey:
Looks good 👍

PR #794, opened 2016-09-23
Added typings for microsoftgraph/msgraph-typescript-typings

dan-silver:
**Typings URL:** https://github.com/microsoftgraph/msgraph-typescript-typings

> I'm not sure this is applicable in our case since the typings are for the Microsoft Graph entities that are returned when hitting the API. This is not typings for a particular JavaScript library.

n/a


unional:
Need to convert from global to module format.
Like this: https://github.com/typings/typings/blob/master/docs/examples.md#named-export

Remove the `declare module` and export directly at top-level

dan-silver:
Thanks @unional! I've converted the typings by removing the global module and exporting all the types and interfaces at the top level.

PR #796, opened 2016-09-24
Fixed filename for Microsoft Graph API typings - wrong GitHub repo was linked

dan-silver:
**Typings URL:** https://github.com/microsoftgraph/msgraph-typescript-typings/blob/master/microsoft-graph.d.ts

We have a [github repo for the Graph typings](https://github.com/microsoftgraph/msgraph-typescript-typings) (objects that are sent to/from the API) and this repo name should be the filename of the registry json file.  I incorrectly linked to our [separate SDK repo](https://github.com/microsoftgraph/msgraph-sdk-javascript) which has methods for interacting with the API.

### DIFF
--- a/common/microsoft-graph.json
+++ b/common/microsoft-graph.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.2.0": "github:microsoftgraph/msgraph-typescript-typings#d964a91520daf8bebc86730a9d9489217d6e8442"
+  }
+}
--- a/npm/msgraph-sdk-javascript.json
+++ b/npm/msgraph-sdk-javascript.json
@@ -1,5 +1,0 @@
-{
-  "versions": {
-    "0.2.0": "github:microsoftgraph/msgraph-typescript-typings#a5bb450b61d0941f4e1fdbde11564243917e83ed"
-  }
-}


felixfbecker:
The name of the JSON file should match the name of the NPM package, not the typings repo

felixfbecker:
It seems to me there is no actual NPM package for this, correct? Is this just for the API?

dan-silver:
These typings aren't really for an npm package. They're for objects that are sent to/from an api so the typings can be used with any http request library like request or superagent (or optionally our SDK).

dan-silver:
I did publish the typings as a separate npm package (https://www.npmjs.com/package/microsoft-graph-typings), but I don't think that's actually useful.  Should I move these typings to a different folder in the registry so it's clear that they are standalone definitions?

felixfbecker:
In that case it shouldn't be in the npm registry, but under common (as they are still external and not global). The name should then be just be something like `msgraph` (it is clear that they are typings), and that name must match the `name` property in typings.json.

/cc @blakeembrey

dan-silver:
Thanks for the help! I've updated the name in typings.json to be microsoft-graph and moved the registry file to common\microsoft-graph.json

PR #802, opened 2016-09-29
Typings for WebdriverIO

sfali23:
**Typings URL:** https://github.com/sfali23/typings-webdriverio

* [X ] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
* [X ] Are they external or global modules according the source (e.g. see README sources)?

### DIFF
--- a/npm/webdriverio.json
+++ b/npm/webdriverio.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#400d940dd4098893cf62bc6315224c44721dc905"
+  }
+}
--- a/npm/webdriverio.json
+++ b/npm/webdriverio.json
@@ -1,5 +1,6 @@
 {
   "versions": {
-    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#2c2123deedd0ef36b08c9de4cbcd32567ce6cfd3"
+    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#49ae26a9976863a61619b715e49ec2aeefa44257",
+	"4.2.16-w": "github:sfali23/typings-webdriverio/wdio#49ae26a9976863a61619b715e49ec2aeefa44257"
   }
 }
--- a/npm/webdriverio.json
+++ b/npm/webdriverio.json
@@ -1,6 +1,5 @@
 {
   "versions": {
-    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#49ae26a9976863a61619b715e49ec2aeefa44257",
-	"4.2.16-w": "github:sfali23/typings-webdriverio/wdio#49ae26a9976863a61619b715e49ec2aeefa44257"
+    "4.2.16": "github:sfali23/typings-webdriverio/standalone#af9892f5172e14368cfba693bd24051ce0ffe691"
   }
 }
--- a/npm/webdriverio.json
+++ b/npm/webdriverio.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "4.2.16": "github:sfali23/typings-webdriverio/standalone#af9892f5172e14368cfba693bd24051ce0ffe691"
+    "4.2.16": "github:sfali23/typings-webdriverio/standalone#ece757c2965a795b8c3686d5a4c4b26f577ed54d"
   }
 }
--- a/npm/webdriverio.json
+++ b/npm/webdriverio.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#400d940dd4098893cf62bc6315224c44721dc905"
+    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#2c2123deedd0ef36b08c9de4cbcd32567ce6cfd3"
   }
 }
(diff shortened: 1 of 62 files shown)


unional:
It seems like `webdriverio` is a npm package. You may consider moving it to `registry/npm/webdriverio.json` and change your typings to external module format. 🌷

sfali23:
There are two modes of API, `standalone` and `wdio` mode. In `standalone` mode API works asynchronously and user is responsible for creating and destroying the `browser` object. The `wdio` mode works with testing frameworks like `mocha`, `cucumber`, or `jasmine` and API works synchronously, in this mode `wdio` framework is responsible for creating and destroying the `browser` object and exposing a global `browser` object. 

The `standalone` mode can be converted to external module format easily, but `wdio` mode requires to have a global `browser` object.

Alternatively, I can move both APIs a external modules and create a separate global module to expose global `browser` object.

Regards,

Farhan

unional:
With TypeScript 2.0, you can use UMD to expose the `browser` object `export as namespace browser`

unional:
btw, `typings.json/ambient` is deprecated. For global typings, set `typings.json/global` to true.

Also, very nicely done typings. 🌷

sfali23:
So I need to move to external module and use typescript 2.0 UMD to expose `browser`.

Any good reference to read about Typescript 2.0.

unional:
They are a bit scatter around right now as TS2.0 just released. Monitor the handbook and search on typescript github repo.

sfali23:
Hi,

Trying to use Typescript 2.0 to expose `browser`, there are lack of good documentation  and examples, trying following code:

```
import * as model from '../lib/model';

import { Browser } from '../lib/wdio/webdriverio';

export { Browser, model }

export as namespace browser;

```

Not sure whether this is correct way to do, but getting following error:

```
TS1316: Global module exports may only appear at top level.
```

Any idea?

Regards,

Farhan

unional:
Have not seen that error before. @blakeembrey any idea?

blakeembrey:
@unional Sadly UMD support can't be done with Typings because it needs to be at the top-level. There's no workaround for it, if you need it you will need to push to DefinitelyTyped so it uses `node_modules/@types`.

unional:
Sorry @sfali23 , that means it is a bad suggestion.

sfali23:
@unional For some reason code snippet you provided I couldn't make it to work. `browser` object couldn't get resolved.

I tried following code in my `index.d.ts` file:

```
declare global {
    interface Object {
        browser: Browser<any>;
    }
}
```

And in the test files I added following code:

```
import { Browser } from 'webdriverio';
let browser: Browser<any> = global.browser;
```

Now `browser` object is available, even though I don't like extra deceleration in each of my file, but this seem to be working and typings is created as external module.

Regards,

Farhan

sfali23:
Correction `let browser = global.browser;` works as well.

sfali23:
migrated to external module as oppose to global module.

unional:
@blakeembrey any suggestion to get this to merge?

npm/webdriverio.json
```diff
@@ -0,0 +1,6 @@
+{
+  "versions": {
+    "4.2.16-s": "github:sfali23/typings-webdriverio/standalone#49ae26a9976863a61619b715e49ec2aeefa44257",
```
felixfbecker: Are these tags actually used by the NPM package?

sfali23: The NPM package version is `4.2.16` but the there are two way access API, standalone and with wdio test runner, in standalone mode API works asynchronously while in wdio mode API works synchronously.

I exposed API for both modes, in order to differentiate two version I added suffix in the version.

felixfbecker: Imo the version should match the NPM version. @blakeembrey ?

felixfbecker: How are they distinguished in the actual NPM package?!

sfali23: They are not distinguished, but when you use `wdio` the API works in synchronously, I guess using `wdio-sync` package.

blakeembrey:
Sorry @sfali23, this is a tricky one for all of us 😄 Bear with us while it's typed out. My suggestion, from perusing the docs and using it a while back, is to make one `npm/` version and another `env/` version. From my understanding, the standalone version is available using NPM while the test-runner gives us the global version (http://webdriver.io/guide/getstarted/modes.html). Does that sound right to you?

felixfbecker:
@blakeembrey `env` or `global`?

sfali23:
@blakeembrey That's correct, so just to confirm, you are suggesting to have standalone as `npm` while test runner version as `env`, is this correct?

blakeembrey:
Correct 👍 

@felixfbecker I think `env` is good, it's not installed as a global and only available when you run the script with `wdio`. Basically, my differentiation is that right now. A `global` would be something like `jQuery` (or maybe even the standalone `wdio` if it were to be a global), while `env` is something explicitly created by a runtime (E.g. `mocha`, `jasmine`, `wdio`, `node`, `electron`, `atom`).

On a side note, I feel like this'll be a tricky one for TS 2.0 with `@types` as well - I wonder if they have a way to disclose the global and module exports at the same time or separately in different compilations without installing two definitions.

sfali23:
This has been done.

PR #807, opened 2016-10-03
Add typings for stash-rest-api v2.6.0

zergmk2:
**Typings URL:** <https://github.com/zergmk2/typed-stash-rest-api>

`None`

Signed-off-by: Lu_Cao <zergmk2@gmail.com>

### DIFF
--- a/npm/node-rush.json
+++ b/npm/node-rush.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.10.0": "github:types/npm-node-rush#12c42b9f4c9bcc821279cc48ba4f39127648bee7"
+  }
+}


blakeembrey:
The only comment I have is on https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L2. Could you use a version that's provided externally so people don't need to install an extra thing to use these?

felixfbecker:
https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L1
I don't think we should have triple slash references like these in typings

https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L2
This is semantically incorrect, `import * as x from 'y'` means x is a plain object of all exports from module y, not a constructor.

https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L7
Instead of defining a namespace, just export all the members directly - otherwise you won't get IntelliSense when doing a selective import.

zergmk2:
@blakeembrey 
I've added the dependency of promise definition in typings.json.
https://github.com/zergmk2/typed-stash-rest-api/blob/master/typings.json#L8

felixfbecker:
@zergmk2 You should use one from the registry ;)

zergmk2:
@felixfbecker 
Thanks for your review comments, below are my responses:
>1) https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L1
I don't think we should have triple slash references like these in typings.

I've updated the code and removed the triple slash references.
>2)https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L2
This is semantically incorrect, import * as x from 'y' means x is a plain object of all exports from module y, not a constructor.

I don't call the Promise's constructor, I just use an exported interface in this module.

>3)https://github.com/zergmk2/typed-stash-rest-api/blob/master/src/index.d.ts#L7
Instead of defining a namespace, just export all the members directly - otherwise you won't get IntelliSense when doing a selective import.

I've explicitly exported all the class and functions in the namespcae.

zergmk2:
@felixfbecker 
oh, I got it. I will update my code. Thank you.

felixfbecker:
> I've explicitly exported all the class and functions in the namespcae.

You should just remove the `namespace` completely

PR #813, opened 2016-10-04
add parseUri typings

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-parseUri

### DIFF
--- a/npm/parseuri.json
+++ b/npm/parseuri.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.0.4": "github:effervescentia/parseuri#d50d66fecb2adc5ea509b8fca99181bf7639e32c"
+    "0.0.4": "github:effervescentia/typed-parseuri#d50d66fecb2adc5ea509b8fca99181bf7639e32c"
   }
 }
--- a/npm/parseuri.json
+++ b/npm/parseuri.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.0.4": "github:effervescentia/typed-parseuri#d50d66fecb2adc5ea509b8fca99181bf7639e32c"
+    "1.2.3-2": "github:effervescentia/typed-parseUri#a9f7d1913184a7befcbfe8b7749746992c8f9f04"
   }
 }
--- a/npm/parseuri.json
+++ b/npm/parseuri.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.0.4": "github:effervescentia/parseuri#d50d66fecb2adc5ea509b8fca99181bf7639e32c"
+  }
+}


felixfbecker:
You know about native url.parse right? That module looks horrible - no readme, commited node_modules, test runner as dependency...

effervescentia:
@felixfbecker the typings yeoman generator picked up the wrong npm package. Supposed to target https://www.npmjs.com/package/parseUri. And I know about it, but that doesn't mean I can use it in the browser, whereas I *can* use this

felixfbecker:
@effervescentia there is a port of the native url package for browserify ;)https://www.npmjs.com/package/url

(of course I have no problem with merging this)

effervescentia:
@felixfbecker using webpack not browserify

felixfbecker:
Doesn't matter, you can just add that package as a dependency ;)

effervescentia:
😓 foiled at every turn... It's smaller? Justifiable?

felixfbecker:
Well it has _way_ more downloads so I would say it is way more reliable/smaller/tested

effervescentia:
https://www.npmjs.com/package/url has just under twice as many downloads as https://www.npmjs.com/package/parseUri, which solely encapsulates the functionality of parsing a uri, whereas `url` also has logic for building urls, so I would argue it is demonstrably smaller. Not to mention that `parseUri`s download statistics are not insignificant by any measure.

unional:
I think it's merely a personal choice. If the typings looks fine I think we should merge it. @felixfbecker ok with you?

blakeembrey:
I merged since it all looked good and the suggestions here were the same I'd give anyway, it's a pretty barren library - no README, source code links, docs, etc. If it works, great, but I'd be hesitant to tell anyone else to use it too 😄

effervescentia:
@blakeembrey thanks for merging.
However, it looks like somehow it's pointing at https://www.npmjs.com/package/parseuri instead of https://www.npmjs.com/package/parseUri. Is this because of the file naming (`parseuri.json`)? Or something else? I also see https://www.npmjs.com/package/parseuri as the source when I run `typings search parseuri`. The generated typings also appear to use the module name `parseuri`.

blakeembrey:
Wow, that's confusing. Yeah, it should have been uppercased in the filename. You can submit a PR to correct that.

effervescentia:
ah damn -__- okay cool, thanks

unional:
Yeah. Since you use `generator-typings`, that comes from the source package.json`

effervescentia:
@unional can confirm that it was the file naming in `typings/registry` and nothing else (at least, seeing as everything else was correct)

unional:
I see. You are correct. The `typings.json` looks fine. 
https://github.com/effervescentia/typed-parseUri/blob/master/typings.json

effervescentia:
@blakeembrey kind of odd behaviour though, would have assumed (along with @unional) that the `name` field in `typings.json` would be responsible

blakeembrey:
Nope, sorry. The `name` field in `typings.json` was a later addition - mostly just for logging in the end. I believe the whole project still works without a name at all. The API itself doesn't actually load anything from your definition to serve it, it just provides pointers from a -> b.

effervescentia:
ah, fair enough 👍  thanks

PR #821, opened 2016-10-10
Add typings for vandium

miltador:
**Typings URL:** https://github.com/miltador/typed-vandium


blakeembrey:
The primary export for this definition should be a function (see https://github.com/vandium-io/vandium-node/#getting-started). Currently, it is not usable in this way. You may want to add some tests to make sure it's typed correctly. The export type you'll want to use to mirror the `module.exports =` behaviour in CommonJS is `export =` in TypeScript.

miltador:
@blakeembrey, thanks for looking into it!
I fixed the typings and also added some basic tests. It's now ready for review.

PR #828, opened 2016-10-13
Create heracles.json

tpluscode:
**Typings URL:** https://github.com/wikibus/heracles

### DIFF
--- a/npm/heracles.json
+++ b/npm/heracles.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.0": "github:wikibus/heracles#a491108bf05c4d1b2540a03c9c9f2604fd11c9e0"
+    "0.1.0": "github:wikibus/heracles#4d5006bae01146ddca084762c95b82965e9db453"
   }
 }
--- a/npm/heracles.json
+++ b/npm/heracles.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.0": "github:wikibus/heracles#4d5006bae01146ddca084762c95b82965e9db453"
+    "0.1.0": "github:wikibus/heracles#99a47e8f25f7dba4b6b176849661780999eb0b8b"
   }
 }
--- a/npm/heracles.json
+++ b/npm/heracles.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.1.0": "github:wikibus/heracles#22f603d36fec7be4f09923b00b8554051b8d2166"
+  }
+}
--- a/npm/heracles.json
+++ b/npm/heracles.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.0": "github:wikibus/heracles#1994a17de6d63a13e5c22bbe25d0305bf8a56f43"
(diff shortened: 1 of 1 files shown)


tpluscode:
Hm, heracles is not a global module. Why does the build think it is?

felixfbecker:
https://github.com/wikibus/heracles/blob/master/typings.json ambientDependencies is deprecated

> Hm, heracles is not a global module. Why does the build think it is?

no idea

blakeembrey:
Also, as the package author it'd probably be cleaner to publish the definition directly to NPM for your users. See https://github.com/typings/registry#as-package-author. It'll still need to be in external format for that to work, which means you need top-level imports and exports to be used.

felixfbecker:
@blakeembrey It has dependencies though

blakeembrey:
@felixfbecker That's fine. It can still be published to NPM. See https://github.com/blakeembrey/popsicle#typescript.

tpluscode:
I was considering NPM but my main target is JSPM. How will JSPM consumers install my typings?

blakeembrey:
Ah, I see - I think that relies on https://github.com/typings/core/pull/140 /cc @unional

tpluscode:
I'm thinking about publishing directly with JSPM/NPM package and I one thing I'm concerned about is keeping versions of code and typings in sync.

Say I want to update my typings to an existing release. Would I have to push new version to NPM or move the tag? For that reason keeping the typings in separate registry seems more appealing to me

felixfbecker:
@tpluscode If you include it with NPM, you will need to publish a new version on NPM.

unional:
Sorry, didn't have time to work on the test part for the PR. I ran tests on `jspm-config` and it is working fine, but definitely want to test it with `typings` itself.

As for versioning, typings in NPM is just a part of the package. As you make changes to your package, just update the typings accordingly and publish as normal.

tpluscode:
I'd like to publish to registry. Keeping the two in sync won't be such a problem.

Please help me out here though. Why is Travis failing with msg

> TypingsError: Unable to read "src/JsonLdUtil.d.ts" from "heracles"

where I import like `import {JsonLdUtil} from "./JsonLdUtil";`

blakeembrey:
@tpluscode I think it fell back to `index.d.ts` which it's following, but it doesn't support `.ts` files (only naive `.d.ts` files, sorry). You can add a `main` to `typings.json` so it knows where to look.

tpluscode:
Oopsies, actually I did miss imports in the `d.ts`. Travis now passes. Please review

tpluscode:
Also, I'd like to tick the checkboxes above with clear conscience. I don't have an `index.js` but `index.d.ts` works fine. Or should I my built output next to the typings?

And what about the READMEs? I'm assuming it applies to dedicated typings repositories but I'm serving directly from my project, correct?

blakeembrey:
The built output seems fine, it's more relevant to bigger projects where it can be important to support deep requires like `x/y`. If you explicitly don't support that, it's not a big deal where the definition is and I'd make it easier on yourself as a maintainer.

For the README, it's fine too. That's for people who may come across the project on GitHub, it helps those users understand what they're looking at when they might be searching for a definition. It also helps us when we review to find the original project through the links.

PR #832, opened 2016-10-16
Add gulp-util.json

demurgos:
**Typings URL:** https://github.com/demurgos/typed-gulp-util

- [x] Does the README explain the purpose of the typings and have a link to the JavaScript project?
- [x] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
- [x] Are they external or global modules according the source (e.g. see README sources)?

### DIFF
--- a/npm/gulp-util.json
+++ b/npm/gulp-util.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.0.7": "github:demurgos/typed-gulp-util#a30aec22a222acd5f5cce2d0c1f96a51d7d7999e"
+    "3.0.7": "github:demurgos/typed-gulp-util#2f2c434fe5bff69a4bd6dafd7e700534d68d20bf"
   }
 }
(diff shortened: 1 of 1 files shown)


demurgos:
I just read about the [namespace hack](https://github.com/typings/discussions/issues/32). I am using it internally: should I update my definitions ?

demurgos:
Well, I checked some other typings and it looks like they are also using `import * as ...` internally, so I'd like to know if my definitions are valid or not.

felixfbecker:
No, they are semantically incorrect by ES6 spec. It's a hack. For modules that use `export =`, use `import x = require('')`

felixfbecker:
LGTM. One minor note: https://github.com/demurgos/typed-gulp-util/blob/master/index.d.ts#L17
What is the reason to export this as a type? You can easily use `typeof colors` to get the type, but you are breaking the naming convention of uppercase names for types here. Also applies a few lines further down again.

demurgos:
This is a remaining from when I was trying to re-export namespaces. Since there is no `export namespace File = Vinyl`, this was the closest I could get so I just tried it - but since it does not work I used a normal namespace for `File` and forgot to remove it for the others.

I just fixed it at the moment.

felixfbecker:
Oh btw you can also just do `export {colors, date}`

demurgos:
Closing as gulp-util is deprecated

PR #842, opened 2016-10-19
Update node.json

blakeembrey:
https://github.com/types/env-node/commit/374f42213af406d9f76e9181da371df3f9e93903 - Fix the missed `interface` in node 4.0 typings.

### DIFF
--- a/env/node.json
+++ b/env/node.json
@@ -1,8 +1,8 @@
 {
   "versions": {
-    "0.10.0": "github:types/env-node/0.10#07b7628aa9af35b23fba65e263336432c8ef8935",
-    "0.12.0": "github:types/env-node/0.12#07b7628aa9af35b23fba65e263336432c8ef8935",
-    "4.0.0": "github:types/env-node/4.0#07b7628aa9af35b23fba65e263336432c8ef8935",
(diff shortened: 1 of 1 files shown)


blakeembrey:
On a side note, it'd be cool to figure out how to type this better and reduce duplication. I've been adding some types/interfaces from 6.2, 6.3, etc. into 6.0 because typing every minor is impossible. Perhaps we need to try and augment minors from the majors, then find a way to augment majors from previous major baselines?

unional:
Would branching instead of "foldering" is a better solution in this regards?

felixfbecker:
I also think so. Because then if the structure is similar you can just cherry-pick the commit on multiple branches.

I also think we should only type the latest major versions.

blakeembrey:
I killed a couple of the major versions, but I'd like to type anything still used in production. It's not a big deal for me to do that, since I still use 0.10 as a min node version in a lot of places.

felixfbecker:
> Edit: Plus, to be honest, it's pretty cool for reference material. It'd be great it we could someday make this as useful as the node docs themselves.

I know, I am already building docs with typedoc from many definitions and the docs are often better than the real JavaScript documentation just because of the power of interfaces for option parameters. Best example https://typed-sequelize.surge.sh/v4/

PR #848, opened 2016-10-24
Create igniteui.json

dkamburov:
**Typings URL:** https://github.com/IgniteUI/typed-igniteui

- [ ] Does the README explain the purpose of the typings and have a link to the JavaScript project?
- [ ] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
- [ ] Are they external or global modules according the source (e.g. see README sources)?


felixfbecker:
Can't review because the file is 3.4MB 😧 Please consider splitting it up into multiple files. Build is also failing

dkamburov:
Well, there are a lot of typings definitions

felixfbecker:
And those are all defined in one file in the source? Typings should follow the source structure.

blakeembrey:
@dkamburov Wow, that's a big definition - did you generate it in some way? Also, as the module author, have you considered providing the definition as part of the library directly?

dkamburov:
Yes, they are generated and are part of IgniteUI, but we're having troubles including the typings in another project - [https://github.com/IgniteUI/igniteui-angular2](https://github.com/IgniteUI/igniteui-angular2). The reason for adding them into the registry is this [issue](https://github.com/IgniteUI/igniteui-angular2/issues/96). The best solution I came up is to include the typings in the typings/registry and refer them as devDependency.

felixfbecker:
If they are generated from TypeScript source, TypeScript should generate multiple declaration files that follow the source structure

dkamburov:
What's the case if they are not generated from a typescript source?

dkamburov:
That may seem like a strange question, but the code is written in JavaScript and the TypeScript definitions were generated based on the comments describing the types of the properties

felixfbecker:
okay. Would it be a major effort to rewrite that tool so it generates multiple files?

dkamburov:
Yes :(

blakeembrey:
@dkamburov Are the definitions meant to be global? Just checking. 

Aside from that, you could still provide the definition as part of the package but it can't be automatically picked up as a global. Only module definitions can work like that (on `import`). I'd recommend providing it with your package to ease development for yourself (no need to manage this registry) and for other users who might not use Typings. I'd probably add a little blurb to your docs to help TypeScript users instead (which should just be "add a reference to this definition to your `tsconfig.json` file").

dkamburov:
Ok, we'll distribute them as part of the igniteui-angular2 package

PR #866, opened 2016-10-28
Add device.js

barinbritva:
**Typings URL:** https://github.com/barinbritva/typed-device.js
**Source URL:** https://github.com/soenkekluth/device.js

Add the type definition for [device.js](https://github.com/soenkekluth/device.js)

### DIFF
--- a/global/device.js.json
+++ b/global/device.js.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "0.2.7": "github:barinbritva/typed-device.js#e2ee13863bf528fd326964379a8d8b02d3e0911d"
+    }
+}
--- a/global/device.js.json
+++ b/global/device.js.json
@@ -1,5 +1,0 @@
-{
-    "versions": {
-        "0.2.7": "github:barinbritva/typed-device.js#16a6c7a5ceeb1eb855831da01fec73684f13fba3"
-    }
-}
--- a/global/device.js.json
+++ b/global/device.js.json
@@ -1,5 +1,5 @@
 {
     "versions": {
-        "0.2.7": "github:barinbritva/typed-device.js#e2ee13863bf528fd326964379a8d8b02d3e0911d"
+        "0.2.7": "github:barinbritva/typed-device.js#16a6c7a5ceeb1eb855831da01fec73684f13fba3"
     }
 }
--- a/npm/device.js.json
+++ b/npm/device.js.json
@@ -1,0 +1,5 @@
+{
+    "versions": {
+        "0.2.7": "github:barinbritva/typed-device.js#3657898d37efa9bc9800572f18a9af865a0d0e52"
+    }
+}


felixfbecker:
Please link the definitions

barinbritva:
@felixfbecker, what do you mean?

felixfbecker:
like mentioned in the PR template, provide a link to the typings repository so we can easily review.

barinbritva:
@felixfbecker, edited.

felixfbecker:
@blakeembrey we have another UMD case here, what was the best way to handle this?

barinbritva:
Damn... It is this project: https://github.com/soenkekluth/device.js. I will update PR.

barinbritva:
Add commit.

blakeembrey:
It looks like this definition is written for the module (e.g. NPM) since the global namespace is incorrect. Should this be written in external module format instead? 

For TypeScript UMD modules, the only thing I think we can do is to recommend TypeScript `@types` because Typings can't handle the simultaneous formats.

barinbritva:
Unforunately, I could not understand: external, internal modules, namespace declaration, module declaration in Typescript. Too complicated for me. I just wrote definition by analogy with other definitions, and it works in my project. I will be very glad, if you will advise me how to write it correct. May be exists an article, which helps understand how to write correct definitions for "typings" definition manager.

felixfbecker:
"external" module format is a file with top-level `import`/`export` statements, like this:
```ts
import {readFile} from 'fs';
export class MyClass {
}
```
The older module declaration format was wrapped in a `declare module "module-name"`:
```ts
declare module "module-name" {
  import {readFile} from 'fs';
  export class MyClass {
  }
}
```
This must not be used with `typings`.
Even older definitions declared a global object, still used sometimes with libraries for the browser:
```ts
declare namespace jQuery {
   export ...
}
```
I would like to help you but as I said I can't even view the file on GitHub because it is so large.

unional:
See: https://github.com/unional/typescript-guidelines/blob/master/pages/typings/namespaces-and-modules.md

barinbritva:
So: if library support umd loading, definition will be something like this:
`
interface Device {
    // interface description
}

// this will work for: import device = require('device.js')
// or for: import device from 'device.js'
declare module 'device.js' {
    export = Device;
}
// will work for including via script tag
declare var device: Device;
`

That is right or not?
I looking for https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/jquery/index.d.ts

barinbritva:
@felixfbecker, what large file do you mean?

felixfbecker:
Ah sorry confused this with another PR

blakeembrey:
@barinbritva How do you use this definition? I assume you use it by doing `import`? If you look at https://github.com/barinbritva/typed-device.js/blob/master/index.d.ts#L6, the global this library introduces is `device.js` whereas the original library introduces only `device` as a global (https://github.com/soenkekluth/device.js/blob/master/device.js#L266).

barinbritva:
@blakeembrey, there is only one problem? If I rename namespace everything will be ok?
How about version which I wrote above?

blakeembrey:
@barinbritva I still need you to help me and answer the questions I asked. How are you using the definition? Are you using it with an `import`? Once I know that, I can help you refactor the definition. I believe you would be using the `import` since you said the definition worked and if you used it the other way, it wouldn't be working. In which case, I'd write the definition in external modules format like this:

```ts
function device(userAgent?: string, classContainer?: Element): device.Device;

declare namespace device {
    export interface Device {
        mobile: () => boolean;
        tablet: () => boolean;
        desktop: () => boolean;
        ios: () => boolean;
        ipad: () => boolean;
        iphone: () => boolean;
        ipod: () => boolean;
        android: () => boolean;
        androidPhone: () => boolean;
        androidTablet: () => boolean;
        blackberry: () => boolean;
        blackberryPhone: () => boolean;
        blackberryTablet: () => boolean;
        windows: () => boolean;
        windowsPhone: () => boolean;
        windowsTablet: () => boolean;
        fxos: () => boolean;
        fxosPhone: () => boolean;
        fxosTablet: () => boolean;
        meego: () => boolean;
        television: () => boolean;

        landscape: () => boolean;
        portrait: () => boolean;

        noConflict: () => Device;
    }
}

export = device;
```

Using external module format is nicer for everyone, and then you can add it to the `npm` source instead of `global`.

felixfbecker:
I would also use methods in the interface instead of arrow functions

unional:
> I would also use methods in the interface instead of arrow functions

Agree. Arrow function will be registered as property instead of function. It is nicer to use function directly so that IDE such as VSC can render completion correctly.

barinbritva:
I updated PR. And I have a few questions.

How I understood, `declare namespace device` says that library device.js adds global variable `device`. That is right?
But how Typescript does understand that `import device = require('device.js');` need to use this definition?
When definition contained:
```javascritpt
declare module 'device.js'
```
IDE suggested definition by ctr+click:
device-require
But now IDE just references to node_modules.

In definition described namespace and function with same name: `device`. IDE thinks that I exports namespace instead of function:
device-unused
How it will be work?

Now definition looks like this:
```typescript
declare namespace device {
    interface CheckingFunction {
        (): boolean;
    }

    interface NoConflictFunction {
        (): Device;
    }
    
    export interface Device {
        mobile: CheckingFunction;
       // and so on
```
Should I to export `CheckingFunction` and `NoConflictFunction ` interfaces?

felixfbecker:
@barinbritva I don't use WebStorm so I can't answer your question. All I can say is the recommended way to write typings is without a `declare module 'whatever'`, but instead just export statements.
You can declare a namespace and a function with the same name and they will get merged (the exports of the namespace will become properties of the function). And yes, export all interfaces.

blakeembrey:
The job of adding `declare module ''` is for Typings. You can do a `typings bundle` or `typings install device=file:device.js.d.ts` to get the bundled declaration suitable for TypeScript consumption.

barinbritva:
Need I do something else for using this typings?
Command `typings install device.js` works. But `npm install @types/device.js` not. I use second option in my project.

felixfbecker:
@barinbritva For that to work we need https://github.com/Microsoft/types-publisher/issues/4

PR #873, opened 2016-10-30
Create farmhash.json

jemmyw:
**Typings URL:** https://github.com/jemmyw/typed-farmhash
- [*] Does the README explain the purpose of the typings and have a link to the JavaScript project?
- [*] Do the typings follow the source structure (e.g. `index.js` <-> `index.d.ts`)?
- [*] Are they external or global modules according the source (e.g. see README sources)?


blakeembrey:
Could you add a link to the original project in the README? There's also a simpler way to type this, in case you're interested. You can just write `export function hash32(...): string;`.

> Does the README explain the purpose of the typings and have a link to the JavaScript project?

PR #888, opened 2016-11-09
Update flux-standard-action

unional:
**Typings URL:** https://github.com/types/npm-flux-standard-action

Consolidate with DT version.
Lower version requirement.

This will break someone who uses this typing as the interface is renamed from `FluxStandardAction` to `Action`, matching what's on `@types`. But it has to be done someway to avoid overall breakage.

### DIFF
--- a/npm/flux-standard-action.json
+++ b/npm/flux-standard-action.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.5.0": "github:types/npm-flux-standard-action#922ff1efb12bc5db4bc0a6c0db49cc8f547d8afd"
+    "0.5.0": "github:types/npm-flux-standard-action#40b444515e68fd085afdf971c138c1ab6b395458"
   }
 }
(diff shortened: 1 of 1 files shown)


felixfbecker:
> Consolidate with DT version

As someone who converted a lot of typings from DT to the registry I cleaned up a lot of bad symbol names. I am against adapting to DT if it means reducing our code quality.

unional:
In this case I think it is good. The way it is written and designed on DT is better than what I had originally. The only questionable piece is the separation of the `meta` field to its own interface.
I can understand it is done to get around 'generic as any' issue, so I'm ok with that.

I've updated the OP with the url so you can take a look to see if you think is it good or not.

felixfbecker:
I think the change is good, but not "to consolidate with DT". That should not be the motivation. The motivation should be to improve the typings, and I wanted to note that here so this doesn't become a reference case to update all typings here "to consolidate with DT" 😉

unional:
haha, sure. 👍

PR #890, opened 2016-11-10
Update flux-standard-action.json

unional:
**Typings URL:** https://github.com/types/npm-flux-standard-action/commits/master

Use type guard.

@felixfbecker , what do you think? I haven't use type guard with generic so I don't know if this is correct.

Test passed, but I don't think I have the right test for it.


felixfbecker:
LGTM

PR #891, opened 2016-11-10
Create jweixin.json

Emeryao:
**Typings URL:** [https://github.com/Emeryao/typed-jweixin](https://github.com/Emeryao/typed-jweixin)

PS: WeChat (Chinese name Weixin) is a mobile social app mainly used in Mainland China, so the documentation is in Chinese.

### DIFF
--- a/global/jweixin.json
+++ b/global/jweixin.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:Emeryao/typed-jweixin#4bf29367274b11d7cbfee9b8b2de1c4755578c04"
+  }
+}


unional:
I'm suspecting this and the `we-app` should be under `env` instead of `global`.

How to you get access to the library when you write code? Through script tag?

Emeryao:
@unional 
For this one, yes, the jweixin library should be referenced in a script tag. 
But for `we-app` it is accessible globally just like the `chrome` variable when developing a chrome extension

unional:
In that case this should remain in `global` and `we-app` should be in `env`

Emeryao:
@unional 
Thanks for your correction. 👍 
I'll fix that and make another PR 😆

PR #913, opened 2016-11-20
Upgrade to v0.0.15

mythz:
**Typings URL:** https://github.com/ServiceStack/servicestack-client/blob/master/lib/index.d.ts

Added support for HTTP Basic Auth + Raw data responses.

### DIFF
--- a/npm/servicestack-client.json
+++ b/npm/servicestack-client.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.0.10": "github:ServiceStack/servicestack-client#1a3c40c74581682efc3345c73d8a2ec038de46fc"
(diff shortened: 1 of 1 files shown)


mythz:
@blakeembrey Hi do you know how long it takes for this to become live as running `install servicestack-client` is still downloading the older version, i.e:

```
typings install servicestack-client
typings INFO globaldependencies "servicestack-client" lists global dependencies on "es6-shim" and "isomorphic-fetch" that must be installed manually
servicestack-client@0.0.10
`-- (No dependencies)
```

blakeembrey:
It's updated within 10 minutes. It has already been updated, the version you see is from your `typings.json` file.

PR #917, opened 2016-11-29
Create highcharts.json

jessezhang91:
**Typings URL:** https://github.com/Tapad/typed-highcharts/blob/master/highcharts.d.ts

It's pretty much a copy-paste from dt~highcharts.

I wasn't sure how to do the JQuery bindings so I just stripped that section out. Is that possible for non-global modules?

### DIFF
--- a/npm/highcharts.json
+++ b/npm/highcharts.json
@@ -1,0 +1,6 @@
+{
+  "versions": {
+    "5.0.5": "github:Tapad/typed-highcharts#72320305a3bb192d445270e248b38599ea59332b"
+  }
+}
+


felixfbecker:
Please export the interfaces.
I think you can do the jQuery augmentation inside a `declare global { }`

jessezhang91:
Woops, sorry! I'm surprised it compiled on my local project without the exports. Made the requisite changes and added the jQuery augmentation.

felixfbecker:
Great, I am a user of that library, so thank you :)

PR #920, opened 2016-11-30
add cosmiconfig typings

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-cosmiconfig


effervescentia:
@felixfbecker I've update it to use a function instead of an interface, however I don't believe I need to export the contents of the namespace. My IDE was able to pick up on the interfaces without me exporting them

felixfbecker:
Yes, they are implicitly exported in .d.ts files. I just personally prefer explicit exports

PR #922, opened 2016-11-30
add klaw typings

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-klaw


effervescentia:
@felixfbecker updated to use `function` instead of `interface`

PR #926, opened 2016-12-04
Update with PR requested by @unional

shawnlauzon:
**Typings URL:** https://github.com/shawnlauzon/typed-soundcloud

From @unional's PR:

> Refactored the code.
> Also fixed the jspm test. I need to make some changes to generator-typings

### DIFF
--- a/npm/soundcloud.json
+++ b/npm/soundcloud.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.1.2": "github:shawnlauzon/typed-soundcloud#cdc030fa3633b7670e98c5b7d551f864843fd145"
+    "3.1.2": "github:shawnlauzon/typed-soundcloud#7849aab4eaec05ad5dea61a185f2170f2c79188d"
   }
 }
--- a/npm/soundcloud.json
+++ b/npm/soundcloud.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "3.1.2": "github:shawnlauzon/typed-soundcloud#b2fd3b91fc9637973f9d39666fd2036180563680"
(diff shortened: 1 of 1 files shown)


felixfbecker:
Why not do the typings ES6-style?
```ts
import { Promise } from 'es6-promise';
export function initialize(options: SoundCloud.InitializeOptions);
export function connect(options?: SoundCloud.ConnectOptions): Promise<any>;

export function get(path: string, params?: any): Promise<any>;
export function post(path: string, params?: any): Promise<any>;
export function put(path: string, params?: any): Promise<any>;
export function delete(path: string): Promise<any>;

export function upload(options: any): Promise<any>;
export function resolve(url: string): Promise<any>;
export function oEmbed(url: string, params?: any): Promise<any>;

export function stream(trackPath: string, secretToken?: string): Promise<SoundCloud.Player>;

export interface InitializeOptions {
    client_id: string;
    redirect_uri?: string;
    oauth_token?: string;
  }

export interface ConnectOptions {
    client_id: string;
    redirect_uri: string;
    scope?: string;
  }

 export interface Player {
    play();
    pause();
    seek(time: number);
    currentTime(): number;
    setVolume(volume: number);
    getVolume(): number;
    on(event: string, handler: any);
  }
}
```

felixfbecker:
Also, `Player.play()` and others are missing a return type annotation - this will cause compilation errors with `noImplicitAny`. You should add that option to tsconfig.json

unional:
This was done quite a while ago. Yes, `Player.play()` and others can add the return type.
As for ES6 vs CommonJS, I think either is fine, but my general suggestion is to keep the typings the same as what the source is to avoid any interop confusion.

shawnlauzon:
Sounds good. I'll add those return types and add `noImplicitAny` to tsconfig.json. As @unional suggests, I'll leave the others as they are.

PR #929, opened 2016-12-06
Add typings for meteor-simple-schema

niceilm:
**Typings URL:** https://github.com/niceilm/typed-meteor-simple-schema

* [v] Does the README explain the purpose of the typings and have a link to the JavaScript project?
* [v] Are they external or global modules according to the source (e.g. see README sources)?

### DIFF
--- a/global/meteor-astronomy.json
+++ b/global/meteor-astronomy.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "1.0.0": "github:niceilm/typed-meteor-astronomy#af9476cb48df49a00cdb255006b58dddfb234c70"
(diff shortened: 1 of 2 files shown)


blakeembrey:
@felixfbecker I think these are fine. There's not a great story for meteor dependencies yet, so the current implementations have all been put in `global`. If someone does come up with a better approach, especially when it comes to those module names, it'd be good to refactor it then. Let me know if I missed something though.

felixfbecker:
Okay, maybe a new registry for meteor, but given the future is NPM `@types` I don't think it's worth the effort.

PR #950, opened 2017-01-20
Create jdbc.json

NetoriaN:
we need this: https://github.com/CraZySacX/node-jdbc

**Typings URL:** 

### DIFF
--- a/npm/jdbc.json
+++ b/npm/jdbc.json
@@ -1,0 +1,5 @@
+{
+  "versions": {
+    "0.4.0": "github:CraZySacX/node-jdbc#f1461d0a2e061ad72b3c3c3a11cf5e4641a3f6d3"
+  }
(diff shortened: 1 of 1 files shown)


unional:
Seems like your are the owner of the library. In this case, you don't need `typings`. Just expose your typings file through your `package.json`.

e.g.:
```js
{
  "name": "...",
  "main": "...",
  "types": "<path to d.ts file>"
}
```

unional:
Seems like you are pointing to the source library. Is the typings in the source project? In that case, you don't need `typings`. Just expose the typings file through `package.json`.

e.g.:
```js
{
  "name": "...",
  "main": "...",
  "types": "<path to d.ts file>"
}
```

NetoriaN:
You are right!

PR #957, opened 2017-02-01
theon initial typings

effervescentia:
**Typings URL:** https://github.com/effervescentia/typed-theon

### DIFF
--- a/npm/theon.json
+++ b/npm/theon.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.28": "github:effervescentia/typed-theon#54b43ad8410eb46d12b3b4fc7c7ebd8f7169f121"
+    "0.1.28": "github:effervescentia/typed-theon#f974f6aa9fd4382e16bfea959a4fea6d4b13014f"
   }
 }
--- a/npm/theon.json
+++ b/npm/theon.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.1.28": "github:effervescentia/typed-theon#cf8c3fc8beb8d915bc3623ed639e29e499f4d0b3"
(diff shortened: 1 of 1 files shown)


unional:
LGTM

blakeembrey:
You shouldn't need all those JSDocs when there's already type annotations. I see at least on spot where they're inconsistent: https://github.com/effervescentia/typed-theon/blob/54b43ad8410eb46d12b3b4fc7c7ebd8f7169f121/lib/theon.d.ts#L294. Also, it'd probably be good to remove all the template `[description]` blocks.

effervescentia:
@blakeembrey fair enough. I only left them in as they were autogen'd and it seemed like more of a pain to go through and remove them. But you're right, no point leaving in JSDocs comments that will inevitably get out of sync

blakeembrey:
@effervescentia What were they generated from?

effervescentia:
@blakeembrey Well I wanted to have the existing JSDocs comments that were in the repo, as I quickly find this library get's difficult to deal with, lot's of circular references. The params were auto-generated for every `/**` block I opened in atom. Wasn't really thinking about it as I was doing it 😝 very late at night

blakeembrey:
I see. I thought maybe you generated the whole definition with a tool or something 😄  It makes sense to add JSDoc comments for intellisense - makes it 100x easier for people to use in the editor instead of having to refer to documentation.

effervescentia:
@blakeembrey agreed 😄 Does a tool like that exist yet? (One that generates typings from JSDoc comments)

blakeembrey:
You could try https://github.com/Microsoft/dts-gen. I haven't tested it myself, but if it's based on the TypeScript compiler parsing JS, it should work.

PR #964, opened 2017-02-12
Update graphql.json

ashleydw:
**Typings URL:** https://github.com/nitintutlani/typed-graphql/

There's no changelog available from the source repo, but looking at the commit history it is largely bug fixes with some additions.

### DIFF
--- a/npm/graphql.json
+++ b/npm/graphql.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "0.5.0": "github:nitintutlani/typed-graphql#8a80d198bb64cc8319d692c3a93af7ed58ed4cc1"
+    "1.0.2": "github:nitintutlani/typed-graphql#13554b8e1c451d47214bd50ba76a5600b51dd4ef"
   }
 }


blakeembrey:
It looks like that module was changed to be a global definition in https://github.com/nitintutlani/typed-graphql/commit/b773a00569ac28d7d70793d9c46b7950628446c3 for some reason. The commit alludes to "TypeScript 2.0" support, but I'm not sure the author understood what TypeScript 2.0 module resolution with NPM was since the change they made was definitely not required (TypeScript 2.0 just allows the NPM `@types` namespace which still uses external modules, module resolution that they added has existed in TypeScript since when Typings was released - since 1.5).

blakeembrey:
It seems we should be able to bump this commit anytime you'd like now 😄

ashleydw:
I've bumped the PR version up. If you can squash the merge that would save me starting a new PR :)

blakeembrey:
@ashleydw Actually, I was wrong. Sorry, I only looked at the comment and not the code itself - it seems they didn't fix the issue and just moved stuff around instead (https://github.com/nitintutlani/typed-graphql/commit/6e0f3251cd51c52ea90fd691a951bdfa21f7f71a).

PR #980, opened 2017-04-04
Remove outdated Electron typings

miniak:
Electron is working on providing an official electron.d.ts (automatically generated from the API docs) electron/electron#4875. Please raise any concerns / ideas there.

### DIFF
--- a/env/electron.json
+++ b/env/electron.json
@@ -1,6 +1,0 @@
-{
-  "versions": {
-    "0.37.4": "github:types/env-electron#12d50b0040ef06e97b5a0bdc0021a0e10b252b3c",
-    "0.37.6": "github:types/env-electron#33b7814e99a685288e87a8756167d3230acaefeb"
-  }
-}


felixfbecker:
@blakeembrey will this break consumers?

miniak:
@felixfbecker I hope nobody is these typings as they are horribly outdated.

unional:
Will not break consumer as it is per commit. 
Just that user can't install a new copy.

miniak:
I would be great if you could clarify the advantages / disadvantages of this repo over DefinitelyTyped. Is there any reason to keep using this place now that npm supports installing typings?

miniak:
Also, the current situation is confusing for users, as they get this error:
```
typings install electron

typings ERR! message Unable to find "electron" ("npm") in the registry.
typings ERR! message However, we found "electron" for 2 other sources: "dt" and "env"
typings ERR! message You can install these using the "source" option.
...
```

unional:
> Also, the current situation is confusing for users, as they get this error:

This, for `typings` user, should not be confusing. That just means you need to do `typings install env~electron` or `typings install dt~electron` (which, in the case, should be the same as `npm install @types/electron`

blakeembrey:
I merged this for now. Feel free to ping me in the deletion thread or just delete the @types repo whenever. Everyone using the current current versions in the registry will continue working, just no one will be able to find these typings going forward. It will break if we delete the repo itself, though, so we can probably just deprecate it for some time first.

PR #988, opened 2017-05-25
update to latest ldapjs type defs

zelein:
**Typings URL:** https://github.com/types/npm-ldapjs

### DIFF
--- a/npm/ldapjs.json
+++ b/npm/ldapjs.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:types/npm-ldapjs#679663089ebdf59a4194f5e03271465ccbf4fe8b"
+    "1.0.0": "github:types/npm-ldapjs#4f8b8a7be5d422b5bed8ac9b4b5b74f8ce4e5064"
   }
 }
--- a/npm/ldapjs.json
+++ b/npm/ldapjs.json
@@ -1,5 +1,5 @@
 {
   "versions": {
-    "1.0.0": "github:types/npm-ldapjs#4f8b8a7be5d422b5bed8ac9b4b5b74f8ce4e5064"
+    "1.0.0": "github:types/npm-ldapjs#753ba6ae60bda6a28d838d2d43bbf0aa7cb20be1"
   }
 }


blakeembrey:
Seems like there might be a bug in `typings` when resolving a relative link?

zelein:
Hmm. I checked and I can't find where that file is being referenced anywhere. In fact, the path referenced does not exist.

And typings is able to successfully build the project when running

`"typings bundle -o test/bundle.d.ts`

zelein:
`typings` appears to think that the reference to `~ldapjs/lib/server` should be resolved to `lib/server/index`, so it must think server is a folder, when it is in fact a module.

blakeembrey:
I think it's worse that that, it searches for `server.d.ts` and falls back to `server/index.d.ts` to support the node.js-like use-cases. However, the issue appears to be the `..` relative paths aren't going up a directory. It's odd, because I'm sure I have other libraries using this pattern - something must be off somewhere.

zelein:
That's the same conclusion I arrived to as well. It's unable to go up a level. However, these type definitions are doing that in plenty of other places, which makes this error even more bizarre.

blakeembrey:
Enabling debug, I see it landing in `https://raw.githubusercontent.com/types/npm-ldapjs/4f8b8a7be5d422b5bed8ac9b4b5b74f8ce4e5064/lib/messages//message.d.ts` (notice the two slashes, which is why it's messing up going up the directory). Not sure where those extra slashes are from yet.

blakeembrey:
Here we are, try https://github.com/types/npm-ldapjs/blob/4f8b8a7be5d422b5bed8ac9b4b5b74f8ce4e5064/lib/messages/bind_response.d.ts#L1.

blakeembrey:
The long term fix _might_ be for `typings` to normalize these paths beforehand so we don't end up in this case.

zelein:
Ahhh good catch. 

With my limited knowledge of `typings`, it appears that it already accounts for this in some cases as during the bundle process it's able to resolve `../result` to `~ldapjs/lib/messages/result`.

Regardless, thank you for tracking that down. I'm submitted another PR upstream to fix the import.

PR #1000, opened 2019-12-16
Move types back to typed-typings

sandersn:
Fixup for moving the types org back to typed-typings.


sandersn:
@blakeembrey I believe this is the correct change although you might want to look over it.

sandersn:
Build fails and will fail until I move the org back to typed-typings. I'm going to do that and restart the CI build.

